PR #1, opened 2025-06-04
Implements pywrangler CLI tool

dom96:
This PR implements a brand new tool called `pywrangler`, which enables the vendoring of Python packages in a Python Workers project.

Screenshot 2025-06-04 at 18 18 20

The tool currently supports the following commands: `sync`, `dev` and `deploy`. The latter of which call `sync` and then proxy to `wrangler`.

The `sync` command reads a pyproject.toml file from the current working directory, it grabs the dependencies specified in that file and executes commands to install those packages into a local vendor folder. The vendor folder is determined by reading the wrangler file's `main` field. The tool also creates a local `.venv-workers` virtual environment and installs those packages in there.

This is mostly an MVP and there are many other features that we will implement in the future, but I hope this forms a good starting point for this tool.

_This PR was implemented with the help of Windsurf_

### Test Plan

```
$ uv clean cache && uv run pytest
```

### DIFF
--- a/src/pywrangler/cli.py
+++ b/src/pywrangler/cli.py
@@ -6,6 +6,7 @@ import click
 
 from pywrangler.sync import (
     check_pyproject_toml,
+    check_requirements_txt,
     is_sync_needed,
     create_pyodide_venv,
     create_workers_venv,
@@ -107,8 +108,9 @@ def sync_command(force=False):
 
     Also creates a virtual env for Workers that you can use for testing.
     """
-    # Check if pyproject.toml exists
+    # Check if pyproject.toml exists and requirements.txt does not
     check_pyproject_toml()
+    check_requirements_txt()
 
     # Check if sync is needed based on file timestamps
     sync_needed = force or is_sync_needed()
--- a/src/pywrangler/sync.py
+++ b/src/pywrangler/sync.py
@@ -28,6 +28,26 @@ def check_pyproject_toml():
         raise click.exceptions.Exit(code=1)
 
 
+def check_requirements_txt():
+    old_requirements_txt = PROJECT_ROOT / "requirements.txt"
+    if old_requirements_txt.is_file():
+        with open(old_requirements_txt, "r") as f:
+            requirements = f.read().splitlines()
+            logger.warning(
+                "Specifying Python Packages in requirements.txt is no longer supported, please use pyproject.toml instead.\n"
+                + "Put the following in your pyproject.toml to vendor the packages currently in your requirements.txt:"
+            )
+            pyproject_text = "dependencies = [\n"
+            pyproject_text += ",\n".join([f'  "{x}"' for x in requirements])
+            pyproject_text += "\n]"
+            logger.warning(pyproject_text)
+
+        logger.error(
+            f"{old_requirements_txt} exists. Delete the file to continue. Exiting."
+        )
+        raise click.exceptions.Exit(code=1)
+
+
 def create_workers_venv():
     """
     Creates a virtual environment at `VENV_WORKERS_PATH` if it doesn't exist.


danlapid:
Can we get CI config for tests/lints/formats/publishing to pypi etc?

README.md
```diff
@@ -1 +1,29 @@
...
+A set of libraries and tools for Python Workers.
+
+
+## Pywrangler 
+
+A CLI tool for managing vendored packages in a Python Workers project.
+
+### Installation
+
+TODO
```
danlapid: How do I install?
Asking for real, I want to try it out

README.md
```diff
@@ -1 +1,29 @@
...
+```bash
+uvx pywrangler --help
+uvx pywrangler sync
+```
+
+
+## Tests
+
+```
+$ uv cache clean
```
danlapid, line 64: Is cache clean required? why?

README.md
```diff
@@ -1 +1,29 @@
...
+
+### Installation
+
+TODO
+
+### Usage
+
+```bash
+uvx pywrangler --help
+uvx pywrangler sync
```
danlapid: Can I then deploy and dev as well?

README.md
```diff
@@ -1 +1,29 @@
...
+```
+
```
danlapid, line 58: What about my IDE? Where should I point it for the lsp to work?

src/pywrangler/cli.py
```diff
@@ -0,0 +1,122 @@
...
+)
+@click.argument("args", nargs=-1, type=click.UNPROCESSED)
+@click.pass_context
+def deploy_command(ctx, args):
+    ctx.invoke(sync_command, force=False)
+    _proxy_to_wrangler("deploy", list(args))
+
+
+if __name__ == "__main__":
+    app()
```
danlapid: What about an unrecognized command? Can we have some general passthrough?

hoodmane: Normally this would go in a file called `__main__.py`.

tests/test_cli.py
```diff
@@ -0,0 +1,348 @@
...
+
+[project]
+name = "test-project"
+version = "0.1.0"
+description = "Test Project"
+requires-python = ">=3.8"
+dependencies = [
+    {','.join([f'"{dep}"' for dep in dependencies])}
+]
+"""
```
hoodmane: I think it is nicer to use `textwrap.dedent` around all of these multiline strings:
```py
    content = dedent(f"""\
        [build-system]
        requires = ["setuptools>=61.0"]
        build-backend = "setuptools.build_meta"
        ...
        dependencies = [
            {','.join([f'"{dep}"' for dep in dependencies])}
        ]
    """)
```

README.md
dom96: I think you're supposed to be able to install it globally via `sudo uv pip install --system -e .` but my homebrew install breaks that.

So I've just been running it like so:

```
uv run --project $PATH_TO_REPO $PATH_TO_REPO/src/pywrangler/cli.py sync
```

README.md
dom96: because `uv` caches files aggressively, I've found that making changes to the Python and rerunning pytest via `uv` doesn't run it against the changes unless you clean the cache

README.md
dom96: yep

README.md
dom96: `./.venv-workers`

Your IDE should hopefully pick that up automatically (it did for me)

src/pywrangler/cli.py
dom96: > Can we have some general passthrough?

You mean all unrecognised commands should go to `wrangler`? That was what I tried initially but Click made it annoyingly difficult to do, so I ended up just defining the commands directly. I can probably perform some checks ahead of Click to do it though.

dom96:
CI implemented, will leave publishing via CI until I can get the access token set up.

src/pywrangler/cli.py
danlapid: Let's find a way to do this, we can't really have most commands fail other than the few we allow.

danlapid:
If all we're doing is calling sync and then proxying to wrangler do we really need click or is it just making things harder than they should be?

README.md
```diff
@@ -1 +1,56 @@
...
+## Pywrangler 
+
+A CLI tool for managing vendored packages in a Python Workers project.
+
+### Installation
+
+You may be able to install the tool globally by running:
+
+```
+uv pip install --system workers-py
```
danlapid, line 16: pip install system fails on macos
```
❯ uv pip install --system workers-py

Using Python 3.13.3 environment at: /opt/homebrew/opt/python@3.13/Frameworks/Python.framework/Versions/3.13
error: The interpreter at /opt/homebrew/opt/python@3.13/Frameworks/Python.framework/Versions/3.13 is externally managed, and indicates the following:

  To install Python packages system-wide, try brew install
  xyz, where xyz is the package you are trying to
  install.

  If you wish to install a Python library that isn't in Homebrew,
  use a virtual environment:

  python3 -m venv path/to/venv
  source path/to/venv/bin/activate
  python3 -m pip install xyz

  If you wish to install a Python application that isn't in Homebrew,
  it may be easiest to use 'pipx install xyz', which will manage a
  virtual environment for you. You can install pipx with

  brew install pipx

  You may restore the old behavior of pip by passing
  the '--break-system-packages' flag to pip, or by adding
  'break-system-packages = true' to your pip.conf file. The latter
  will permanently disable this error.

  If you disable this error, we STRONGLY recommend that you additionally
  pass the '--user' flag to pip, or set 'user = true' in your pip.conf
  file. Failure to do this can result in a broken Homebrew installation.

  Read more about this behavior here: <https://peps.python.org/pep-0668/>

Consider creating a virtual environment with `uv venv`.
```

danlapid:
`uv run pywrangler --help` shows very few commands, we should make sure to show the entire wrangler help.
I wonder if we can just proxy everything more easily rather than the click based solution, then proxy help as well?

Maybe intercept help to add `sync` though.

README.md
dom96: Yeah, I run into the same thing on macOS hence why I wrote "may" in the text above, but I can make it more explicit.

README.md
danlapid, line 1: Do I still need
```
    "rules": [
        {
            "globs": [
                "vendor/**/"
            ],
            "type": "Data",
            "fallthrough": true
        }
    ],
```
in my wrangler.jsonc?
Can we get around that?

src/pywrangler/cli.py
danlapid, line 1: uv run pywrangler --help shows very few commands, we should make sure to show the entire wrangler help.
I wonder if we can just proxy everything more easily rather than the click based solution, then proxy help as well?

Maybe intercept help to add sync though.

README.md
danlapid: Yeah let's do that,
Also the following works really well:

Edit your project's pyproject.toml with:
```
...
[dependency-groups]
dev = ["workers-py"]

[tool.uv.sources]
workers-py = { path = "../workers-py" }
```
then run
```
uv run pywrangler ...
```

src/pywrangler/cli.py
danlapid, line 1: If all we're doing is calling sync and then proxying to wrangler do we really need click or is it just making things harder than they should be?

README.md
dom96: Have you tried it without those rules? You shouldn't need it.

danlapid: I think I hallucinated, it works now.

src/pywrangler/cli.py
dom96: Perhaps. But I do think we want our own --help, we can append wrangler's help to it, but things like `pywrangler sync --help` should work. We also may add more commands in the future so having a good basis is nice.

README.md
danlapid: Ok I see what's going on, it's only passing python files so I get 
```
ModuleNotFoundError: No module named 'pydantic_core._pydantic_core'
```

danlapid, line 1: When using this it seems to also vendor workers-py as vendor/workers.py which causes conflict to import `from workers import DurableObject, Response, handler`
We should make sure it doesn't vendor workers-py itself

danlapid: Fixed with dev dependencies, we should make sure to note that!!!
```
[project]
name = "python-mcp"
version = "0.1.0"
requires-python = "==3.12"
dependencies = ["mcp", "structlog"]

[dependency-groups]
dev = ["workers-py"]

[tool.uv.sources]
workers-py = { path = "../workers-py" }

```

src/pywrangler/cli.py
danlapid: Ok, either works for me but we should definitely support all of wrangler's commands.

src/pywrangler/__init__.py
danlapid, line 1: How do I specify which wrangler version to use?

src/pywrangler/cli.py
```diff
@@ -0,0 +1,119 @@
...
+        logger.warning(
+            "No dependencies found in [project.dependencies] section of pyproject.toml."
+        )
+    install_requirements()
+
+    write_success("Sync process completed successfully.")
+
+
+def _proxy_to_wrangler(command_name, args_list):
+    command_to_run = ["npx", "wrangler", command_name] + args_list
```
danlapid: Can we get a TODO somewhere or a JIRA ticket to track 
1. using user-specified wrangler version
2. not requiring a node installation?

README.md
dom96: hah that's fun, I'm impressed that it just works (TM) with `tool.uv.sources`

src/pywrangler/sync.py
```diff
@@ -0,0 +1,267 @@
...
+            pass
+
+    # Add webtypy to the venv requirements file for type hints
+    with open(VENV_REQUIREMENTS_PATH, "r") as f:
+        current_content = f.read()
+
+    with open(VENV_REQUIREMENTS_PATH, "a") as f:
+        if current_content and not current_content.endswith("\n"):
+            f.write("\n")
+        f.write("webtypy\n")
```
danlapid: @hoodmane should we add "pyodide-py"?

src/pywrangler/sync.py
```diff
@@ -0,0 +1,267 @@
...
+
+    # Install packages into .venv-workers so that user's IDE can see the packages.
```
danlapid: Can this be extracted to a function?

danlapid: In general a few more sub-functions here would help readability.

README.md
hoodmane: Presumably we can teach wrangler to apply these settings by default for Python packages.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,30 @@
...
+      run: |
+        pip install uv
+        
+    - name: Lint with Ruff
+      run: |
+        uvx ruff check .
+        
+    - name: Format with Ruff
+      run: |
+        uvx ruff format --check .
```
hoodmane: Missing end of file newline.

hoodmane: I'd also suggest applying ruff in pre-commit, but that can be applied in a followup.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,39 @@
...
+      run: |
+        uv pip install --system -e .
+        
+    - name: Run tests
+      run: |
+        uv run pytest
+
+    - name: Verify that pywrangler can be run globally
+      run: |
+        pywrangler --help
```
hoodmane: Newline at end of file

pyproject.toml
```diff
@@ -1,7 +1,59 @@
+[build-system]
+requires = ["setuptools>=61.0"]
+build-backend = "setuptools.build_meta"
```
hoodmane, line 3: I'd suggest switching to hatchling. But we can do that in a followup. To use hatchling, we need this diff to pyproject.toml:
```patch
diff --git a/pyproject.toml b/pyproject.toml
index 46ed2d1..8a321e6 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,10 +1,9 @@
 [build-system]
-requires = ["setuptools>=61.0"]
-build-backend = "setuptools.build_meta"
+requires = ["hatchling", "hatch-vcs"]
+build-backend = "hatchling.build"
 
 [project]
 name = "workers-py"
-version = "0.1.0"
 description = "A set of libraries and tools for Python Workers"
 readme = "README.md"
 requires-python = ">=3.10"
@@ -19,6 +18,7 @@ dependencies = [
     "commentjson>=0.9.0",
     "pyodide-cli",
 ]
+dynamic = ["version"]
 
 [dependency-groups]
 dev = [
@@ -57,3 +57,9 @@ no_implicit_optional = true
 module = "tests.*"
 disallow_untyped_defs = false
 disallow_incomplete_defs = false
+
+[tool.hatch.version]
+source = "vcs"
+
+[tool.hatch.build.targets.wheel]
+packages = ["src/pywrangler"]
```

src/pywrangler/cli.py
```diff
@@ -0,0 +1,119 @@
...
+    logger.info(f"Passing command to npx wrangler: {' '.join(command_to_run)}")
+    try:
+        process = subprocess.run(command_to_run, check=False, cwd=".")
+        click.get_current_context().exit(process.returncode)
+    except FileNotFoundError:
+        logger.error(
+            "'npx' or 'wrangler' not found. Ensure Node.js and Wrangler are installed and in your PATH."
+        )
```
hoodmane: Should we also log the original error message here? It would probably contain more specific information about what is missing.

src/pywrangler/sync.py
```diff
@@ -0,0 +1,267 @@
...
+
+try:
+    import tomllib  # Standard in Python 3.11+
+except ImportError:
+    import tomli as tomllib  # For Python < 3.11
+
+logger = logging.getLogger(__name__)
+
+# Define paths
+PROJECT_ROOT = Path.cwd()  # Assumes script is run from project root
```
hoodmane, line 18: Maybe search up until we find a `pyproject.toml` so we can run from subdirectory? You could use a variant of this function here:
https://github.com/pyodide/pyodide-build/blob/main/pyodide_build/common.py#L556-L579

src/pywrangler/sync.py
```diff
@@ -0,0 +1,267 @@
...
+
+def create_workers_venv():
+    """
+    Creates a virtual environment at `VENV_WORKERS_PATH` if it doesn't exist.
+    """
+    if not VENV_WORKERS_PATH.is_dir():  # Check if it's a directory
+        logger.debug(f"Creating virtual environment at {VENV_WORKERS_PATH}...")
+        run_command(["uv", "venv", str(VENV_WORKERS_PATH), "--python", "python3.12"])
+    else:
+        logger.debug(f"Virtual environment at {VENV_WORKERS_PATH} already exists.")
```
hoodmane: Nit: avoid the negative:
```suggestion
    if VENV_WORKERS_PATH.is_dir():  # Check if it's a directory
        logger.debug(f"Virtual environment at {VENV_WORKERS_PATH} already exists.")
        return
    logger.debug(f"Creating virtual environment at {VENV_WORKERS_PATH}...")
    run_command(["uv", "venv", str(VENV_WORKERS_PATH), "--python", "python3.12"])
```

src/pywrangler/sync.py
```diff
@@ -0,0 +1,267 @@
...
+def _get_pyodide_cli_path():
+    venv_bin_path = VENV_WORKERS_PATH / ("Scripts" if os.name == "nt" else "bin")
+    pyodide_cli_path = venv_bin_path / ("pyodide.exe" if os.name == "nt" else "pyodide")
+    return pyodide_cli_path
+
+
+def install_pyodide_build():
+    pyodide_cli_path = _get_pyodide_cli_path()
+
+    if not pyodide_cli_path.is_file():
```
hoodmane: Similarly I think early return is nicer:
```
    if pyodide_cli_path.is_file():
        logger.debug(
            f"pyodide-build CLI already found at {pyodide_cli_path} (skipping install.)"
        )
        return

File: src/pywrangler/sync.py [hunk 18]
```diff
@@ -0,0 +1,267 @@
...
+        dependencies = pyproject_data.get("project", {}).get("dependencies", [])
+
+        if not dependencies:
+            return False
+
+        # Write dependencies to requirements.txt
+        GENERATED_REQUIREMENTS_PATH.parent.mkdir(parents=True, exist_ok=True)
+        with open(GENERATED_REQUIREMENTS_PATH, "w") as req_file:
+            for dep in dependencies:
+                req_file.write(f"{dep}\n")
```
hoodmane: Slightly shorter:
```suggestion
        GENERATED_REQUIREMENTS_PATH.write_text("\n".join(dependencies))
```

src/pywrangler/sync.py
```diff
@@ -0,0 +1,267 @@
...
+    run_command([str(pyodide_cli_path), "venv", str(PYODIDE_VENV_PATH)])
+
+
+def generate_requirements() -> bool:
+    logger.debug(
+        f"Reading dependencies from {PYPROJECT_TOML_PATH} and generating {GENERATED_REQUIREMENTS_PATH}..."
+    )
+    try:
+        with open(PYPROJECT_TOML_PATH, "rb") as f:  # Use binary mode for tomllib
+            pyproject_data = tomllib.load(f)
```
hoodmane: Let's make the try/except here a bit more targeted:
```suggestion
    try:
        with open(PYPROJECT_TOML_PATH, "rb") as f:
            pyproject_data = tomllib.load(f)
    except  tomllib.TOMLDecodeError as e:
        logger.error(f"Error parsing {PYPROJECT_TOML_PATH}: {str(e)}")
        raise click.exceptions.Exit(code=1)
```

src/pywrangler/sync.py
```diff
@@ -0,0 +1,267 @@
...
+        raise click.exceptions.Exit(code=1)
+
+    return True
+
+
+def install_requirements():
+    # Get the vendor path dynamically from wrangler config
+    try:
+        vendor_path_relative = get_vendor_path_from_wrangler_config(PROJECT_ROOT)
+    except Exception as e:
```
hoodmane: More specific:
```suggestion
    except (ValueError, FileNotFoundError) as e:
```

src/pywrangler/sync.py
```diff
@@ -0,0 +1,267 @@
...
+            extra={"markup": True},
+        )
+
+    # Clean up temporary files
+    if GENERATED_REQUIREMENTS_PATH.exists():
+        GENERATED_REQUIREMENTS_PATH.unlink()
+        logger.debug(f"Cleaned up {GENERATED_REQUIREMENTS_PATH}.")
+    if VENV_REQUIREMENTS_PATH.exists():
+        VENV_REQUIREMENTS_PATH.unlink()
+        logger.debug(f"Cleaned up {VENV_REQUIREMENTS_PATH}.")
```
hoodmane: If these are temporary, maybe put them in the temp directory them rather than in the project directory?

src/pywrangler/sync.py
```diff
@@ -0,0 +1,267 @@
...
+
+
+def check_timestamps():
```
hoodmane: Maybe a better name would be:
```suggestion
def is_sync_needed():
```

src/pywrangler/sync.py
```diff
@@ -0,0 +1,267 @@
...
+    try:
+        vendor_path_relative = get_vendor_path_from_wrangler_config(PROJECT_ROOT)
+        vendor_path = PROJECT_ROOT / vendor_path_relative
+
+        if vendor_path.is_dir():
+            vendor_mtime = vendor_path.stat().st_mtime
+            vendor_needs_update = pyproject_mtime > vendor_mtime
+    except Exception:
+        # If we can't determine the vendor path, default to requiring an update
+        vendor_needs_update = True
```
hoodmane, line 262:
```suggestion
        return True
```

src/pywrangler/sync.py
```diff
@@ -0,0 +1,267 @@
...
+        # If pyproject.toml doesn't exist, we need to abort anyway
+        return True
+
+    pyproject_mtime = PYPROJECT_TOML_PATH.stat().st_mtime
+
+    # Check if .venv-workers exists and get its timestamp
+    venv_needs_update = True
+    if VENV_WORKERS_PATH.is_dir():
+        venv_mtime = VENV_WORKERS_PATH.stat().st_mtime
+        venv_needs_update = pyproject_mtime > venv_mtime
```
hoodmane:
```suggestion
    if not VENV_WORKERS_PATH.is_dir():
        return True

    venv_mtime = VENV_WORKERS_PATH.stat().st_mtime
    venv_needs_update = pyproject_mtime > venv_mtime
    if venv_needs_update:
        return True
```

src/pywrangler/sync.py
hoodmane:
```suggestion
    try:
        vendor_path_relative = get_vendor_path_from_wrangler_config(PROJECT_ROOT)
    except (ValueError, FileNotFoundError):
        # If we can't determine the vendor path, default to requiring an update
        return True
    vendor_path = PROJECT_ROOT / vendor_path_relative
    if not vendor_path.is_dir():
        return True
    vendor_mtime = vendor_path.stat().st_mtime
    vendor_needs_update = pyproject_mtime > vendor_mtime
    return vendor_needs_update
```

src/pywrangler/sync.py
```diff
@@ -0,0 +1,267 @@
...
+        ) as dest:
+            dest.write(src.read())
+    else:
+        # Create a new empty requirements file
+        with open(VENV_REQUIREMENTS_PATH, "w") as f:
+            pass
+
+    # Add webtypy to the venv requirements file for type hints
+    with open(VENV_REQUIREMENTS_PATH, "r") as f:
+        current_content = f.read()
```
hoodmane, line 180:
```suggestion
    current_content = VENV_REQUIREMENTS_PATH.read_text()
```

src/pywrangler/sync.py
```diff
@@ -0,0 +1,267 @@
...
+
```
hoodmane, line 181:
```suggestion
```

src/pywrangler/sync.py
```diff
@@ -0,0 +1,267 @@
...
+def generate_requirements() -> bool:
+    logger.debug(
+        f"Reading dependencies from {PYPROJECT_TOML_PATH} and generating {GENERATED_REQUIREMENTS_PATH}..."
+    )
+    try:
+        with open(PYPROJECT_TOML_PATH, "rb") as f:  # Use binary mode for tomllib
+            pyproject_data = tomllib.load(f)
+
+        # Extract dependencies from [project.dependencies]
+        dependencies = pyproject_data.get("project", {}).get("dependencies", [])
```
hoodmane, line 118: How about just returning the dependencies here? And then we can put `@functools.cache` around this? No need to create a file.

src/pywrangler/sync.py
```diff
@@ -0,0 +1,267 @@
...
+            f"Pyodide virtual environment at {PYODIDE_VENV_PATH} already exists."
+        )
+        return
+
+    logger.debug(f"Creating Pyodide virtual environment at {PYODIDE_VENV_PATH}...")
+    PYODIDE_VENV_PATH.parent.mkdir(parents=True, exist_ok=True)
+    run_command([str(pyodide_cli_path), "venv", str(PYODIDE_VENV_PATH)])
+
+
+def generate_requirements() -> bool:
```
hoodmane: How about:
```suggestion
def get_requirements() -> list[str]:
```

src/pywrangler/utils.py
```diff
@@ -0,0 +1,155 @@
...
+    logging.addLevelName(RUNNING_LEVEL, "RUNNING")
+    logging.addLevelName(OUTPUT_LEVEL, "OUTPUT")
+
+
+def write_success(msg):
+    logging.log(SUCCESS_LEVEL, msg)
+
+
+def run_command(
+    command: list[str],
```
hoodmane:
```suggestion
    command: list[str | Path],
```

src/pywrangler/sync.py
```diff
@@ -0,0 +1,267 @@
...
+        )
+        run_command(
+            [
+                "uv",
+                "pip",
+                "install",
+                "-p",
+                str(venv_python_executable),
+                "-r",
+                str(VENV_REQUIREMENTS_PATH),
```
hoodmane:
```suggestion
                venv_python_executable,
                "-r",
                VENV_REQUIREMENTS_PATH,
```

src/pywrangler/cli.py
dom96: Done in commit that implements proxying for all commands.

src/pywrangler/cli.py
dom96: EW-9300 and EW-9301

src/pywrangler/sync.py
dom96: Sure, but will do in a follow up.

README.md
dom96: > Presumably we can teach wrangler to apply these settings by default for Python packages.

Tracked in EW-9302

src/pywrangler/cli.py
```diff
@@ -0,0 +1,151 @@
...
+                timeout=10,
+            )
+            if result.returncode == 0:
+                wrangler_help = result.stdout
+                # Replace 'wrangler' with 'pywrangler' in the help text
+                wrangler_help = wrangler_help.replace("wrangler ", "pywrangler ")
+                # Indent each line of the wrangler help
+                indented_help = "\n".join(
+                    "  " + line for line in wrangler_help.split("\n")
+                )
```
hoodmane: Use `textwrap.indent()` for this.

dom96:
All feedback resolved, should be good for another pass.

danlapid (approved):

PR #12, opened 2025-06-20
fix: use jsonc-parser for jsonc parsing to support multi-line comments

dom96:


hoodmane (approved):
Thanks!

danlapid (approved):
LGTM but pyjson5 seems to be the most maintained and most used library for jsonc files

PR #15, opened 2025-07-10
feat: renames vendor dir to `python_modules`

dom96:
This changes where we put our vendored python modules. Now that we always put the vendored files in a `python_modules` directory, we no longer need to parse the wrangler.toml file, so it removes a bunch of code as well.

Related to https://github.com/cloudflare/workerd/pull/4493.

### Test Plan

```
$ uv cache clean && uv run pytest
```


danlapid (approved):
Approving, but there's still some references to "vendor" in the code, can we do a find all and edit?

src/pywrangler/sync.py
```diff
@@ -46,6 +45,16 @@ def check_requirements_txt():
         raise click.exceptions.Exit(code=1)
 
 
+def check_wrangler_config():
+    wrangler_jsonc = PROJECT_ROOT / "wrangler.jsonc"
+    wrangler_toml = PROJECT_ROOT / "wrangler.toml"
+    if not wrangler_jsonc.is_file() and not wrangler_toml.is_file():
```
peturingi, line 51: Reduce negation complexity:
```suggestion
    if not (wrangler_jsonc.is_file() or wrangler_toml.is_file()):
```

tests/test_cli.py
```diff
@@ -183,7 +183,7 @@ def test_sync_command_integration(dependencies, clean_test_dir):
             assert all(
                 d.name.startswith("__") for d in TEST_SRC_VENDOR.iterdir() if d.is_dir()
             ), (
-                f"Vendor directory should be empty of packages but contains: {list(TEST_SRC_VENDOR.iterdir())}"
+                f"python_modules directory should be empty of packages but contains: {list(TEST_SRC_VENDOR.iterdir())}"
```
peturingi, line 186:
```suggestion
                f"python_modules directory must be empty of packages but contains: {list(TEST_SRC_VENDOR.iterdir())}"
```
[RFC-2119](https://datatracker.ietf.org/doc/html/rfc2119)

peturingi (approved):

PR #17, opened 2025-07-23
fix: output relative path in package installation message

dom96:
Nice and easy one.

Before:

```
INFO     Installing packages into /Users/dominik/temp/onsite-demo/python_modules using Pyodide pip...
```

After:

```
INFO     Installing packages into python_modules using Pyodide pip...
```


hoodmane (approved):
Thanks!

PR #20, opened 2025-08-01
Pass --yes to npx so it won't time out after 10 seconds if wrangler not installed

hoodmane:
Maybe we should also reduce the timeout time?


dom96 (approved):
Yep, reducing the time out time sounds good

PR #22, opened 2025-08-06
fix: Allow overriding the python version

hoodmane:


dom96 (approved):

dom96:
I guess we'll need to implement auto detection of this at some point, perhaps based on the compat flags/date in the wrangler file.

PR #24, opened 2025-08-22
fix: lock pyodide-build to fix running on Py 3.12

dom96:
Right now pywrangler is broken when run with the Python version set to 3.12 (the default). This PR fixes this.

This PR also changes the env var to just contain the version, rather than the `python` prefix too.

### DIFF
--- a/src/pywrangler/sync.py
+++ b/src/pywrangler/sync.py
@@ -55,6 +55,10 @@ def check_wrangler_config():
         raise click.exceptions.Exit(code=1)
 
 
+def _get_python_version():
+    return os.environ.get("_PYWRANGLER_PYTHON_VERSION", "3.12")
+
+
 def create_workers_venv():
     """
     Creates a virtual environment at `VENV_WORKERS_PATH` if it doesn't exist.
@@ -64,8 +68,10 @@ def create_workers_venv():
         return
 
     logger.debug(f"Creating virtual environment at {VENV_WORKERS_PATH}...")
-    python = os.environ.get("_PYWRANGLER_PYTHON_VERSION", "python3.12")
-    run_command(["uv", "venv", str(VENV_WORKERS_PATH), "--python", python])
+    python_version = _get_python_version()
+    run_command(
+        ["uv", "venv", str(VENV_WORKERS_PATH), "--python", f"python{python_version}"]
+    )
 
 
 def _get_pyodide_cli_path():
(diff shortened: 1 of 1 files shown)


src/pywrangler/sync.py
```diff
@@ -98,8 +104,21 @@ def install_pyodide_build():
 
     run_command(["uv", "pip", "install", "-p", str(venv_python_executable), "pip"])
 
+    pyodide_build_requirement = (
```
hoodmane: Doesn't look right to me. If this is necessary you open a bug report against pyodide-build with a reproducer and link it here. But I think the newer version is compatible, we have test coverage for this in pyodide-build and cibuildwheel.

dom96: https://github.com/pyodide/pyodide-build/issues/238

dom96: I think there is definitely a breakage here as our CI is failing with the latest pyodide-build. At the very least we should lock the version of pyodide-build we install to make sure this doesn't happen again.

dom96: Okay, turns out our CI wasn't failing. But there is some strange caching going on on my machine that triggers this. Hood suggested running `./.venv-workers/bin/pyodide xbuildenv install` as a workaround and that seems to work.

src/pywrangler/sync.py
```diff
@@ -111,6 +126,10 @@ def create_pyodide_venv():
         )
         return
 
+    # Workaround to fix caching issue on some machines.
+    logger.debug("Installing xbuildenv...")
+    run_command([str(pyodide_cli_path), "xbuildenv", "install"])
```
hoodmane: Add a link to the PR fixing it:
https://github.com/pyodide/pyodide-build/pull/239

src/pywrangler/sync.py
```diff
@@ -98,8 +104,17 @@ def install_pyodide_build():
 
     run_command(["uv", "pip", "install", "-p", str(venv_python_executable), "pip"])
 
+    pyodide_build_requirement = "pyodide-build==0.30.7"
     run_command(
```
hoodmane, line 107: How about just:
```
run_command(["uv", "pip", "install", "-p", str(venv_python_executable), "pip", "pyodide-build==0.30.7"])
```
We don't need `"--prerelease=allow"` for `0.30.7` I don't think.

hoodmane (approved):

PR #30, opened 2025-09-25
feat Use uv instead of pyodide-build to manage pyodide install and venv

hoodmane:


src/pywrangler/sync.py
```diff
@@ -216,32 +186,39 @@ def _install_requirements_to_vendor(requirements: list[str]):
     with tempfile.NamedTemporaryFile(
         mode="w", suffix=".txt", dir=PYODIDE_VENV_PATH
     ) as temp_file:
+        print("requirements", requirements)
```
dom96:
```suggestion
```

src/pywrangler/sync.py
```diff
@@ -216,32 +186,39 @@ def _install_requirements_to_vendor(requirements: list[str]):
...
+                "-p",
+                interp_name,
+                "--no-build",
                 "-r",
                 str(temp_file_path),
-            ]
+                "--extra-index-url",
+                "https://index.pyodide.org/0.27.7",
+                "--index-strategy",
+                "unsafe-best-match"
```
dom96: will this unsafe-best-match index strategy cause us any problems? Just wondering why it's called "unsafe"

src/pywrangler/sync.py
```diff
@@ -216,32 +186,39 @@ def _install_requirements_to_vendor(requirements: list[str]):
...
+            ],
+            env=os.environ | {"VIRTUAL_ENV" :  str(PYODIDE_VENV_PATH)}
         )
+        pyv = _get_python_version()
+        shutil.rmtree(vendor_path)
+        shutil.copytree(PYODIDE_VENV_PATH / f"lib/python{pyv}/site-packages", vendor_path)
```
dom96:
```suggestion
        site_packages_path = PYODIDE_VENV_PATH / f"lib/python{pyv}/site-packages"
        logger.debug(f"Copying {site_packages_path} to {vendor_path}")
        shutil.copytree(site_packages_path, vendor_path)
```

src/pywrangler/sync.py
```diff
@@ -253,61 +230,7 @@ def _install_requirements_to_vendor(requirements: list[str]):
         )
 
 
-def _install_requirements_to_venv(requirements: list[str]):
-    # Create a requirements file for .venv-workers that includes webtypy and pyodide-py
-    VENV_REQUIREMENTS_PATH.parent.mkdir(parents=True, exist_ok=True)
-
-    requirements = requirements.copy()
-    requirements.append("webtypy")
-    requirements.append("pyodide-py")
```
dom96, line 277: Don't see these being added anywhere anymore. Are these being installed implicitly some other way?

dom96 (approved):
Looks good, but possibly need to ensure we install webtypy/pyodide-py

src/pywrangler/sync.py
hoodmane: We need it so that if one version of a package is in the pyodide index but that version does not satisfy requirements, it will still fall back and use a different version from pypi. This is the closest setting to what PIP does:

> While unsafe-best-match is the closest to pip's behavior, it exposes users to the risk of "dependency confusion" attacks.

Otherwise, we fail to resolve in cases where we previously succeeded. I think the reason it's called unsafe is for internal indices:
1. private org index contains package foo
2. attacker publishes a package called foo to pypi with more versions
3. someone in the private org accidentally caps foo to a version that is not actually present, private org accidentally gets unrelated attacker foo package from pypyi

This threat model doesn't make sense for us because the packages in the Pyodide index are also in PyPI and maintained by more or less trusted OSS maintainers.

PR #35, opened 2025-10-07
feat: implements python version detection based on wrangler config

dom96:
* reads the compat flag/compat date
* computes the python version based on that information

This is a fairly straightforward way to get this information. In the future we can consider exposing this data from workerd/wrangler and querying it in pywrangler to avoid duplicating the metadata here.


src/pywrangler/sync.py
```diff
@@ -60,15 +63,118 @@ def check_wrangler_config():
...
+    """
+    wrangler_toml = PROJECT_ROOT / "wrangler.toml"
+    wrangler_jsonc = PROJECT_ROOT / "wrangler.jsonc"
+
+    if wrangler_toml.is_file():
+        try:
+            with open(wrangler_toml, "rb") as f:
+                return tomllib.load(f)
+        except tomllib.TOMLDecodeError as e:
+            logger.warning(f"Error parsing {wrangler_toml}: {e}")
```
hoodmane: Isn't this likely to be unrecoverable?

src/pywrangler/sync.py
```diff
@@ -60,15 +63,118 @@ def check_wrangler_config():
...
+    # If no config was found, return None
+    if not config:
+        return None
+
+    compat_flags = config.get("compatibility_flags", [])
+    compat_date = config.get("compatibility_date")
+
+    # Check if python_workers base flag is present (required for Python workers)
+    if "python_workers" not in compat_flags:
+        logger.warning("`python_workers` compat flag not specified in wrangler config")
```
hoodmane: Again we could probably fail here since eventually it's not going to work.

src/pywrangler/sync.py
```diff
@@ -60,15 +63,118 @@ def check_wrangler_config():
...
+        return None
+
+    # Find the most specific Python version based on compat flags and date
+    # Sort by version descending to prioritize newer versions
+    sorted_versions = sorted(
+        PYTHON_COMPAT_VERSIONS, key=lambda x: x.version, reverse=True
```
hoodmane, line 130: This is a constant why are we sorting it? Can't we just create it ordered in the right way?

src/pywrangler/sync.py
```diff
@@ -60,15 +63,118 @@ def check_wrangler_config():
...
+    for py_version in sorted_versions:
+        # Check if the specific compat flag is present
+        if py_version.compat_flag in compat_flags:
+            return py_version.version
+
+        # For versions with compat_date, also check the date requirement
+        if py_version.compat_date:
+            if compat_date:
+                try:
+                    config_date = datetime.strptime(compat_date, "%Y-%m-%d")
```
hoodmane: Surely we should strptime the compat_date immediately after reading it?

src/pywrangler/sync.py
```diff
@@ -60,15 +63,118 @@ def check_wrangler_config():
...
+                    cutoff_date = datetime.strptime(py_version.compat_date, "%Y-%m-%d")
```
hoodmane: This one is from a global variable maybe we should store the datetime in the global rather than parsing it here?

src/pywrangler/sync.py
```diff
@@ -60,15 +63,118 @@ def check_wrangler_config():
...
+    Priority order:
+    1. Environment variable _PYWRANGLER_PYTHON_VERSION
+    2. Wrangler config compat flags and compat date
+    3. Default to 3.12
+
+    Returns:
+        Python version string ("3.12" or "3.13")
+    """
+    # Check environment variable first
+    env_version = os.environ.get("_PYWRANGLER_PYTHON_VERSION")
```
hoodmane: We could maybe just get rid of this environment variable.

src/pywrangler/sync.py
```diff
@@ -60,15 +63,118 @@ def check_wrangler_config():
...
+    # TODO: In the future we should query `workerd` for the Python version being used so we
+    # have a single source of truth for this information.
+    config_version = _get_python_version_from_wrangler_config()
+    if config_version:
+        logger.debug(f"Using Python version from wrangler config: {config_version}")
+        return config_version
+
+    # Default fallback
+    logger.debug("Using default Python version: 3.12")
+    return "3.12"
```
hoodmane: What situations would we hit this? When there is no wrangler config? I wonder if this should also be an error.

README.md
```diff
@@ -58,7 +58,14 @@ workers-py = { path = "../workers-py" }
 
 Then run via `uv run pywrangler`.
 
-## Tests
+#### Lint
+
+```
+uv run ruff check --fix
+uv run ruff format
```
hoodmane, line 65:
```suggestion
uv run pre-commit
```

tests/test_py_version_detect.py
```diff
@@ -0,0 +1,169 @@
...
+
+class TestPythonVersionDetection(unittest.TestCase):
+    def setUp(self):
+        self.temp_dir = tempfile.mkdtemp()
+        self.temp_path = Path(self.temp_dir)
+
+    def tearDown(self):
+        import shutil
+
+        shutil.rmtree(self.temp_dir)
```
hoodmane: Just use the `tmp_path` fixture.

tests/test_py_version_detect.py
```diff
@@ -0,0 +1,169 @@
...
+import tempfile
+import unittest
+from pathlib import Path
+from unittest.mock import patch
+
+# Import the functions we want to test
+import pywrangler.sync as sync_module
+
+
+class TestPythonVersionDetection(unittest.TestCase):
```
hoodmane: I think we should remove the class and move the tests to top level.

tests/test_py_version_detect.py
```diff
@@ -0,0 +1,169 @@
...
+        self.temp_path = Path(self.temp_dir)
+
+    def tearDown(self):
+        import shutil
+
+        shutil.rmtree(self.temp_dir)
+
+    def test_wrangler_toml_with_compat_flag(self):
+        """Test Python 3.13 detection with python_workers_20250116 compat flag."""
+        with patch.object(sync_module, "PROJECT_ROOT", self.temp_path):
```
hoodmane: Instead, use the monkeypatch fixture

tests/test_py_version_detect.py
```diff
@@ -0,0 +1,169 @@
+"""Tests for sync module Python version detection."""
+
+import os
+import tempfile
+import unittest
```
hoodmane: Maybe we need to add a CLAUDE.md file to tell claude that we use pytest for testing?

src/pywrangler/sync.py
dom96: Sure, but if we make a mistake and fail to sort it correctly this logic will fail. It's better to ensure it is sorted at the point where the code requires it to be. The alternative is to check if it's sorted correctly and fail, I think this is preferable.

tests/test_py_version_detect.py
dom96: Yeah, good idea. I'll add it.

README.md
dom96: hmm, pre-commit didn't run these commands

```
check for added large files..........................(no files to check)Skipped
check for case conflicts.............................(no files to check)Skipped
check for merge conflicts............................(no files to check)Skipped
check for broken symlinks............................(no files to check)Skipped
check yaml...........................................(no files to check)Skipped
debug statements (python)............................(no files to check)Skipped
fix end of files.....................................(no files to check)Skipped
mixed line ending....................................(no files to check)Skipped
trim trailing whitespace.............................(no files to check)Skipped
fix requirements.txt.................................(no files to check)Skipped
ruff.................................................(no files to check)Skipped
ruff-format..........................................(no files to check)Skipped
shellcheck...........................................(no files to check)Skipped
codespell............................................(no files to check)Skipped
```

hoodmane: It only runs on files with staged changes. If you want to run them all do:
```
uv run pre-commit run -a
```

src/pywrangler/sync.py
```diff
@@ -60,15 +63,98 @@ def check_wrangler_config():
...
+
+    for py_version in sorted_versions:
+        # Check if the specific compat flag is present
+        if py_version.compat_flag in compat_flags:
+            return py_version.version
+
+        # For versions with compat_date, also check the date requirement
+        if py_version.compat_date:
+            if compat_date:
+                if compat_date >= py_version.compat_date:
```
hoodmane: Can we use `and` here rather than three nested ifs?

src/pywrangler/sync.py
```diff
@@ -60,15 +63,98 @@ def check_wrangler_config():
...
-            raise ValueError(
-                f"Unexpected value for Python version '{res}', expected '3.12' or '3.13'"
-            )
+    """
+    Determine the Python version to use based on wrangler configuration.
+
+    Returns:
+        Python version string
+    """
+    config_version = _get_python_version_from_wrangler_config()
```
hoodmane: Let's just combine this function with `_get_python_version_from_wrangler_config()` since they are the same?

hoodmane (approved):
Generally looks good to me, left a couple more nits.

README.md
dom96: I see, well we may as well document the ruff commands anyway

PR #36, opened 2025-10-07
fix: Add version check for uv

hoodmane:


dom96 (approved):
Awesome, thanks for adding this. Will it make it friendlier to figure out what went wrong for anyone that runs into this case.

PR #38, opened 2025-10-09
feat: implement pywrangler types to generate Python type stubs

hoodmane:
Example use:
```toml
kv_namespaces = [
    { binding = "FOO", id = "<YOUR_KV_NAMESPACE_ID>" }
]
```
Then `uv run pywrangler types`. Then we get type hints:
```py
from workers import WorkerEntrypoint, Response
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from js import Env

    env: "Env"
        reveal_type(self.env.FOO) # Type of "self.env.FOO" is "KVNamespace_iface"
        bar = await self.env.FOO.get("bar")
        reveal_type(bar) # Type of "bar" is "str | None"
        return Response(bar) 
```


src/pywrangler/cli.py
```diff
@@ -90,6 +88,25 @@ def app(ctx, debug=False):
         logger.setLevel(logging.DEBUG)
 
 
+@app.command("types")
+@click.option(
+    "-o",
+    "--outdir",
+    type=click.Path(writable=True),
+    help="The output directory to write the generated types. Default: .venv/lib/python3.vv/site-packages/js-stubs",
```
dom96, line 96: wrangler defaults to saving them in ./worker-configuration.d.ts

https://developers.cloudflare.com/workers/wrangler/commands/#types

Should we do something similar and encourage the user to commit it to their repo?

src/pywrangler/cli.py
```diff
@@ -90,6 +88,25 @@ def app(ctx, debug=False):
...
+)
+@click.option(
+    "-c",
+    "--config",
+    type=click.Path(exists=True, dir_okay=False, readable=True),
+    help="Path to Wrangler configuration file",
+)
```
dom96, line 103: You should detect this automatically, it'll either be a wrangler.toml or a wrangler.jsonc file

src/pywrangler/types.py
```diff
@@ -0,0 +1,61 @@
...
+        outdir = find_outdir()
+    if config:
+        args += ["--config", config]
+    with TemporaryDirectory() as tmp_str:
+        tmp = Path(tmp_str)
+        run_command(WRANGLER_COMMAND + args + [tmp / "worker-configuration.d.ts"])
+        (tmp / "tsconfig.json").write_text(TSCONFIG)
+        (tmp / "package.json").write_text(PACKAGE_JSON)
+        run_command(["npm", "-C", tmp, "install"])
+        run_command(["npx", "@pyodide/ts-to-python", tmp, outdir / "__init__.pyi"])
```
dom96: This means we'll have another dependency on `npm`. That's probably fine but something to keep in mind as we pursue removing this dependency.

src/pywrangler/cli.py
hoodmane: It's detected automatically but you can also pass it explicitly. This matches the behavior of `npx wrangler types` which also allows passing an explicit config path.

hoodmane: This is optional, if the user doesn't pass it it will be detected automatically.

PR #52, opened 2025-10-21
chore: enable more mypy checks

hoodmane:


pyproject.toml
```diff
@@ -64,14 +64,25 @@ lint.flake8-comprehensions.allow-dict-calls-with-keyword-arguments = true
...
 check_untyped_defs = true
-disallow_any_generics = false
+disallow_any_generics = true
 no_implicit_optional = true
+warn_redundant_casts = true
+warn_unused_ignores = true
+no_implicit_reexport = true
+strict_equality = true
+disallow_subclassing_any = true
+disallow_untyped_calls = true
```
dom96: why do all of these need to be enabled one by one? Isn't there some sort of "strict" mode that enables these and other common ones all at once?

dom96 (approved):

pyproject.toml
hoodmane: There is a strict mode that enables many but not all of these.

PR #53, opened 2025-10-21
feat: install dynamic library dependencies for scipy if present

hoodmane:
This installs libopenblas.so if scipy is installed. It's a bit ad-hoc for now. Eventually we should fix it for all packages with dynamic library deps.

- [ ] Add a test

### DIFF
--- a/src/pywrangler/sync.py
+++ b/src/pywrangler/sync.py
@@ -1,3 +1,4 @@
+import json
 import logging
 import os
 import shutil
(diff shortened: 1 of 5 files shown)


src/pywrangler/sync.py
```diff
@@ -239,9 +248,49 @@ def _install_requirements_to_venv(requirements: list[str]) -> None:
...
+    with lock_path.open() as f:
+        res = json.load(f)
+    return cast(dict[str, PyodidePackageMetadata], res["packages"])
+
+
+def _maybe_install_dylibs() -> None:
+    vendor_path = get_vendor_path()
+    if not (vendor_path / "scipy").exists():
+        return
+    libdir = vendor_path / "lib"
```
dom96, line 275:
```suggestion
    libdir = vendor_path / "lib"
    logger.info("Installing dylibs for scipy")
```

PR #54, opened 2025-10-30
feat: better errors when unsupported packages are requested

dom96:
* output is now captured for uv commands
* packages are now installed to venv first
* this enables us to determine the cause of installation errors more accurately
* error messages are now clear about the cause of the failure and let the user know if a package is unsupported by python workers

### Test Plan

* with package version that works in native python but not in pyodide

```
$ uv tool install -e . # install dev copy of pywrangler
# in wrangler project with "dspy==3.0.3" in pyproject.toml
$ pywrangler sync 
INFO     Found 2 dependencies.                                                                                                                                                                                                               
INFO     Installing packages into .venv-workers...                                                                                                                                                                                           
INFO     Packages installed in .venv-workers.                                                                                                                                                                                                
INFO     Installing packages into python_modules...                                                                                                                                                                                          
WARNING  Using Python 3.13.2 environment at: .venv-workers/pyodide-venv                                                                                                                                                                      
           × No solution found when resolving dependencies:
....
ERROR    Installation failed because the packages you requested are not supported by Python Workers. See above for details.
```

* with package version that doesn't exist at all

```
# in wrangler project with "dspy==4.0.3" in pyproject.toml
$ pywrangler sync   
INFO     Found 2 dependencies.                                                                                                                                                                                                               
INFO     Installing packages into .venv-workers...                                                                                                                                                                                           
WARNING  Using Python 3.13.1 environment at: .venv-workers                                                                                                                                                                                   
           × No solution found when resolving dependencies:                                                                                                                                                                                  
           ╰─▶ Because there is no version of dspy==4.0.3 and you require dspy==4.0.3, we can conclude that your requirements are unsatisfiable.                                                                                             
ERROR    Failed to install the requirements defined in your pyproject.toml file. See above for details.
```

### DIFF
--- a/src/pywrangler/sync.py
+++ b/src/pywrangler/sync.py
@@ -177,7 +177,7 @@ def _install_requirements_to_vendor(requirements: list[str]) -> None:
         extra={"markup": True},
     )
     with temp_requirements_file(requirements) as requirements_file:
-        run_command(
+        result = run_command(
             [
                 "uv",
                 "pip",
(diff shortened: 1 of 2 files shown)


src/pywrangler/utils.py
```diff
@@ -79,20 +79,30 @@ def run_command(
     """
     logger.log(RUNNING_LEVEL, f"{' '.join(str(arg) for arg in command)}")
     try:
```
hoodmane, line 81: Maybe:
```py
kwargs = {}
if capture_output:
    kwargs = dict(
        stdout=subprocess.PIPE,
        stderr=subprocess.STDOUT
    )
```
and then add `**kwargs` to the `subprocess.run()` call.

src/pywrangler/sync.py
```diff
@@ -240,8 +271,14 @@ def _install_requirements_to_venv(requirements: list[str]) -> None:
 
 
 def install_requirements(requirements: list[str]) -> None:
-    _install_requirements_to_vendor(requirements)
```
hoodmane, line 243: Maybe we should rename these two functions to include `pyodide` in one and `native` in the other, since it's a bit vague. Also, we should first install the packages to the pyodide venv and then lock the same versions for the native venv. If the pyodide installation fails, in the error path we can try installing it natively and decide what error message to display based on whether that works or not.

dom96: Sure. But I think we should do that separately. This PR represents a good QoL improvement for the error messages (and we don't perform the locking as things stand, I have created EW-9718 already to track this).

src/pywrangler/utils.py
dom96: I tried this with `{"stdout": subprocess.PIPE, "stderr": subprocess.STDOUT}` but mypy didn't like it. I guess `dict` may work though?

hoodmane: I think it'll cause the same problem but you can just put a `# type: ignore`.

hoodmane: I think a `type: ignore` is better than duplicating the call.

PR #56, opened 2025-11-25
chore: remove unnecessary `exclude` from pre-commit config

ryanking13:
I think these are copied from pyodide, which is not needed in this repository.


ryanking13:
Hmm I cannot reproduce the macos error locally. It looks like the type generation is somehow working incorrectly.

@hoodmane Do you have any idea?

PR #59, opened 2026-01-19
fix: ensure the same version is installed in host and pyodide venv

ryanking13:
This ensures the same version is installed in the host venv and Pyodide venv.

The solution is a little bit ugly...

1. Install packages from PyPI
2. Install packages from Pyodide index
3. (new) Re install packages from PyPI with the same versions in Pyodide index

Once we have PEP 783 accepted and all the packages are hosted in PyPI, we should be able to get rid of these steps.

---

Before this change, the packages installed in the two venv were different, as follows:

```
FAILED tests/test_cli.py::test_sync_command_integration[dependencies1] - AssertionError: Version mismatch for numpy: vendor has 2.0.2, venv has 2.4.1
```

### DIFF
--- a/tests/test_version_sync.py
+++ b/tests/test_version_sync.py
@@ -1,5 +1,4 @@
-from pathlib import Path
-from unittest.mock import Mock, patch
+from unittest.mock import patch
 
 import pywrangler.sync as pywrangler_sync
 
@@ -22,6 +21,7 @@ def test_parse_pip_freeze():
     )
 
     assert result == ["shapely==2.0.7", "numpy==1.26.4"]
+
 
 class TestInstallRequirements:
     @patch.object(pywrangler_sync, "_install_requirements_to_vendor")
@@ -54,11 +54,7 @@ class TestInstallRequirements:
             i for i, msg in enumerate(log_messages) if mocked_native_error in msg
         )
         pyodide_idx = next(
-            (
-                i
-                for i, msg in enumerate(log_messages)
-                if mocked_pyodide_error in msg
-            ),
+            (i for i, msg in enumerate(log_messages) if mocked_pyodide_error in msg),
             None,
         )
         assert pyodide_idx is None, (
@@ -95,7 +91,11 @@ class TestInstallRequirements:
 
         log_messages = [record.message for record in caplog.records]
         assert any(mocked_pyodide_error in msg for msg in log_messages)
-        assert any("Installation of packages into the Python Worker failed. Possibly because these packages are not currently supported. See above for details." in msg for msg in log_messages)
+        assert any(
+            "Installation of packages into the Python Worker failed. Possibly because these packages are not currently supported. See above for details."
+            in msg
+            for msg in log_messages
+        )
 
     @patch.object(pywrangler_sync, "_install_requirements_to_vendor")
     @patch.object(pywrangler_sync, "_get_vendor_package_versions")


src/pywrangler/sync.py
```diff
@@ -229,56 +220,123 @@ def _install_requirements_to_vendor(requirements: list[str]) -> None:
...
+    accurately reflect what will run in production.
+
+    Returns:
+        Error message string if installation failed, None if successful.
+    """
+    requirements = requirements + ["pyodide-py"]
+    error = _pip_install_to_venv(
+        requirements,
+        get_venv_workers_path(),
+        action_name="Installing packages into",
```
dom96: Hmm, this is only used here. So why the refactoring? Also it seems like the changes made to the code that used to be in `_install_requirements_to_venv` aren't necessary. The changes seem to mostly be stylistic as far as I can tell. Would be nice to revert this to make it easier to see what the functional changes are here and to go back to the previous version which is simpler.

src/pywrangler/sync.py
```diff
@@ -229,56 +220,123 @@ def _install_requirements_to_vendor(requirements: list[str]) -> None:
...
+
+    # Then install to .venv-workers using the pinned versions from vendor.
+    # This ensures host packages accurately reflect what will run in production.
+    native_error = _install_requirements_to_venv(_get_vendor_package_versions())
+
+    # Show the native error first (more likely to be actionable), then the Pyodide error.
+    if native_error:
+        logger.warning(native_error)
+        logger.error(
+            "Failed to install the requirements defined in your pyproject.toml file. See above for details."
```
dom96, line 327: This is no longer true I guess. You now install packages based on the pinned versions in vendor, so I suppose this be changed to reflect this.

src/pywrangler/sync.py
```diff
@@ -229,56 +220,123 @@ def _install_requirements_to_vendor(requirements: list[str]) -> None:
...
+                "Installation failed because of an invalid peer certificate. Are your systems certificates correctly installed? Do you have an Enterprise VPN enabled?"
+            )
+        elif "failed to fetch" in lowered_error:
+            logger.error(
+                "Installation failed because of a failed fetch. Is your network connection working?"
+            )
+        elif "no solution found when resolving dependencies" in lowered_error:
+            logger.error(
+                "Installation failed because the packages you requested are not supported by Python Workers. See above for details."
+            )
```
dom96, line 346: Since you've changed the order of operations here, I'm worried that users will no longer get this nice error message when they request an unsupported package. Will they still get this message?

src/pywrangler/sync.py
ryanking13: Good catch. I extracted it in my previous version which installed packages 3 times, then updated it after addressing Hood's comment. Now it is used only once and this refactoring is not very useful.

src/pywrangler/sync.py
ryanking13: Hmm, but it is still true that we are installing packages in pyproject.toml, and I am not sure what is a more intuitive error message for this.

If the installation fails here, I think it is most likely that a user passed an invalid package name or version that is not possible to be installed in native environments anyways.

src/pywrangler/sync.py
ryanking13: While the order of operations are changed, the order of error messages are the same. We show the native error first, and then the Pyodide error. So there should be no changes in what error message the user is getting.

src/pywrangler/sync.py
dom96: Okay, fair enough. I think the issue I had in mind doesn't apply now that I thought it through in more detail.

src/pywrangler/sync.py
dom96: Hmm, well, I think I need to be clearer about the problem here. Here is the old code (read the comments):

```python
    # Note: the order these are executed is important.
    # We need to install to .venv-workers first, so that we can determine if the packages requested
    # by the user are valid.
    _install_requirements_to_venv(requirements)
    # Then we install the same requirements to the vendor directory. If this installation
    # fails while the above succeeded, it implies that Pyodide does not support these package
    # requirements which allows us to give a nicer error message to the user.
    _install_requirements_to_vendor(requirements)
```

But you've swapped these. So we can no longer distinguish between Python packages that aren't supported by Pyodide only vs. Python packages that aren't supported by anything.

So with your new ordering, the user will get this error message even if they ask for packages that also aren't supported by Python itself. In other words: the error message loses its accuracy. It would be nice if we could still detect this, but if not then I think we should at least change the error message wording.

ryanking13: I added three different test cases which are

1) package isn't supported by any platform (native error should be shown)
2) package isn't supported by Pyodide only (pyodide error should be shown)
3) package is supported in Pyodide but not in native platform (this should be extremely rare, but just in case).

dom96: Hmm, none of your tests seem to assert that this "not supported by Python Workers" error is logged. Did I miss something? If not can you add a test which does so?

ryanking13: Oh, right. I added testcase for each common cases as well.

dom96: Thanks!

tests/test_version_sync.py
```diff
@@ -0,0 +1,215 @@
+from pathlib import Path
```
dom96: In general there is a lot of reliance for mocks in these tests. I wonder if we could avoid the mocks to get better test coverage. I left some suggestions below.

tests/test_version_sync.py
```diff
@@ -0,0 +1,215 @@
...
+        mock_vendor_path.return_value = Path("/fake/python_modules")
+        mock_pyodide_path.return_value = Path("/fake/pyodide-venv")
+        mock_run_command.return_value = Mock(
+            returncode=0,
+            stdout="shapely==2.0.7\nnumpy==1.26.4\nclick==8.1.7\n",
+        )
+
+        result = pywrangler_sync._get_vendor_package_versions()
+
+        assert result == ["shapely==2.0.7", "numpy==1.26.4", "click==8.1.7"]
```
dom96: This is basically testing the parsing implemented in `_get_vendor_package_versions`. Instead of mocking, can you refactor the parsing logic into its own function and call that directly to verify it works as intended? This way all the separate tests below become one test with multiple asserts, simplifying the testing logic a lot.

tests/test_version_sync.py
```diff
@@ -0,0 +1,215 @@
...
+        call_order = []
+        mock_vendor.side_effect = lambda r: call_order.append("vendor")
+        mock_get_vendor.side_effect = lambda: ["shapely==2.0.7", "numpy==1.26.4"]
+        mock_venv.side_effect = lambda r: call_order.append("venv")
+
+        pywrangler_sync.install_requirements(["click", "numpy"])
+
+        assert call_order == ["vendor", "venv"]
+        mock_vendor.assert_called_once_with(["click", "numpy"])
+        mock_venv.assert_called_once_with(["shapely==2.0.7", "numpy==1.26.4"])
```
dom96: This is just an implementation detail, we probably don't need a test for it, do we? if we do can you leave a comment explaining it?

tests/test_version_sync.py
```diff
@@ -0,0 +1,215 @@
...
+        assert mock_get_vendor.call_count == 0
+
+        assert mock_vendor.call_args_list[0][0][0] == ["some-package"]
+
+        # native installation should be called with the original requirements
+        assert mock_venv.call_args_list[0][0][0] == ["some-package"]
+
+        log_messages = [record.message for record in caplog.records]
+        assert any("Pyodide install failed" in msg for msg in log_messages)
+        assert any("Python Worker failed" in msg for msg in log_messages)
```
dom96: Can we expand these error messages? I guess this is part of the "Installation of packages into the Python Worker failed." message. We should assert on the full message to avoid false positives.

tests/test_version_sync.py
```diff
@@ -0,0 +1,215 @@
...
+        # Pyodide installation failed, so _get_vendor_package_versions should not be called
+        assert mock_get_vendor.call_count == 0
+
+        assert mock_vendor.call_args_list[0][0][0] == ["some-package"]
+
+        # native installation should be called with the original requirements
+        assert mock_venv.call_args_list[0][0][0] == ["some-package"]
+
+        log_messages = [record.message for record in caplog.records]
+        assert any("Pyodide install failed" in msg for msg in log_messages)
```
dom96:
```suggestion
        assert any(mocked_error in msg for msg in log_messages)
```

tests/test_version_sync.py
```diff
@@ -0,0 +1,215 @@
...
+        )
+        assert native_idx is not None
+
+    @patch.object(pywrangler_sync, "_install_requirements_to_vendor")
+    @patch.object(pywrangler_sync, "_get_vendor_package_versions")
+    @patch.object(pywrangler_sync, "_install_requirements_to_venv")
+    def test_only_pyodide_error_shown_when_native_succeeds(
+        self, mock_venv, mock_get_vendor, mock_vendor, caplog
+    ):
+        mock_vendor.return_value = "Pyodide install failed: no solution found"
```
dom96:
```suggestion
        mocked_error = "Pyodide install failed: no solution found"
        mock_vendor.return_value = mocked_error
```

tests/test_version_sync.py
```diff
@@ -0,0 +1,215 @@
...
+    @patch.object(pywrangler_sync, "_install_requirements_to_vendor")
+    @patch.object(pywrangler_sync, "_get_vendor_package_versions")
+    @patch.object(pywrangler_sync, "_install_requirements_to_venv")
+    def test_known_pyodide_errors(
+        self, mock_venv, mock_get_vendor, mock_vendor, caplog
+    ):
+        common_errors = {
+            "invalid peer certificate": "Are your systems certificates correctly installed? Do you have an Enterprise VPN enabled?",
+            "failed to fetch": "Is your network connection working?",
+            "no solution found when resolving dependencies": "the packages you requested are not supported by Python Workers. See above for details.",
```
dom96, line 141: Any chance we can mock this at a higher level? or better yet run it e2e on a package that is known to be unsupported by Pyodide? I know that this kind of thing can change, but it would be nice to have a full test for this.

tests/test_version_sync.py
ryanking13: Yeah, it is unfortunate these tests rely a lot on mocking, but I am not sure if there is a better way in terms of unittest. Probably having a more robust integration tests would be a good way for us to be certain about these changes.

tests/test_version_sync.py
ryanking13: Sounds good. I extracted pip freeze parsing logic.

tests/test_version_sync.py
ryanking13: Updated error messages.

tests/test_version_sync.py
ryanking13: Makes sense. Removed this test.

tests/test_version_sync.py
ryanking13: Sounds reasonable. Maybe let's create a separate PR that separates the unittest and e2e tests. I think currently they are mixed up in the same file and hard to distinguish.

PR #60, opened 2026-01-19
chore: run type tests only in Python>=3.13

ryanking13:
The typing generated by `ts-to-python` package uses Python>=3.13 syntax (https://peps.python.org/pep-0696/), which is currently making the CI fail

```
FAILED tests/test_types.py::test_types - assert 'Revealed type is "js.Env"' in "src/js-stubs/__init__.pyi:109: error: Invalid syntax. Maybe you meant '==' or ':=' instead of '='?  [syntax]\nFound 1 error in 1 file (errors prevented further checking)\n"
 +  where "src/js-stubs/__init__.pyi:109: error: Invalid syntax. Maybe you meant '==' or ':=' instead of '='?  [syntax]\nFound 1 error in 1 file (errors prevented further checking)\n" = CompletedProcess(args=['uv', 'run', 'mypy'], returncode=2, stdout="src/js-stubs/__init__.pyi:109: error: Invalid synta... the workspace. Defaulting to `>=3.12`.\nDownloading mypy (11.7MiB)\n Downloaded mypy\nInstalled 7 packages in 16ms\n').stdout
```

The problematic syntax is setting the default value for the generic, which is supported only after Python 3.13.

```py
type CfProperties[HostMetadata=Any]
```

Maybe we can improve ts-to-python library to handle this case, but to unblock other pull requests, I changed the test to be only run in python 3.13, and included Python 3.13 in CI

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: ['3.12']
+        python-version: ['3.12', '3.13']
     runs-on: ${{ matrix.os }}
 
     steps:
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -1,5 +1,8 @@
+import sys
 from contextlib import chdir
 from subprocess import run
+
+import pytest
 
 # Import the full module so we can patch constants
 from pywrangler.types import wrangler_types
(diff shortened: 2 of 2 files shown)


hoodmane:
I think the commit message should say "chore" here since otherwise it will make a release.

hoodmane (approved):
Thanks. As I said, please change commit message to say "chore:" instead of "fix:".

ryanking13:
> 
I think the commit message should say "chore" here since otherwise it will make a release.

Oh, I see. Thanks for the heads up. Didn't know we are using commit message-based release.

PR #62, opened 2026-01-28
fix: better windows platform support

ryanking13:
Fixes a few minor errors that I found while working on Windows support

- Passing `Path` object in env paramter seems to error in Windows, explictily converting it to string.
- Ensure `encoding='utf-8'` when running subprocess for machines that do not use utf-8
- `npm` or `npx` can be a batchfile in windows

### DIFF
--- a/src/pywrangler/utils.py
+++ b/src/pywrangler/utils.py
@@ -143,7 +143,7 @@ def run_command(
 
     # Some tools like `npm` may be a batch file on Windows (npm.cmd), and calling them only by
     # name may fails in subprocess.run. Use shutil.which to find the real name.
-    abspath = shutil.which(str(command[0]))
+    abspath = shutil.which(command[0])
     if not abspath:
         logger.error(f"Command not found: {command[0]}. Is it installed and in PATH?")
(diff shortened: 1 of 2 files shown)


src/pywrangler/utils.py
```diff
@@ -138,6 +140,16 @@ def run_command(
         A subprocess.CompletedProcess instance.
     """
     logger.log(RUNNING_LEVEL, f"{' '.join(str(arg) for arg in command)}")
+
+    # Some tools like `npm` may be a batch file on Windows (npm.cmd), and calling them only by
+    # name may fails in subprocess.run. Use shutil.which to find the real name.
+    abspath = shutil.which(str(command[0]))
```
dom96: The type of `command` is `list[str]` so the `str` shouldn't be necessary, right?

```suggestion
    abspath = shutil.which(command[0])
```

dom96 (approved):
Nice! Thank you!

src/pywrangler/utils.py
ryanking13: Good point. Right.

Issue #58, opened 2026-01-14
Support package projects with src-layout


lemonyte:
I have an ASGI app project with the following [src-layout](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/) file structure:

```
project_name
├── config.toml
├── pyproject.toml
├── wrangler.toml
└── src
    ├── package_name
    │   ├── __init__.py
    │   ├── server.py
    │   ├── ...
    │   ├── subpackage_name
    │   │   └── __init__.py
    │   ├── data
    │   │   └── data.json
    │   ├── static
    │   │   └── styles.css
    │   └── templates
    │       └── index.html
    ├── js-stubs
    │   └── __init__.pyi
    └── worker.py
```

During runtime, it needs to read files in subdirectories of the package (e.g. `data`, `templates`), and files in the current working directory (e.g. `config.toml`).

If I put `worker.py` inside the package and import the ASGI app, it fails because `package_name` is not installed as a package, so Python doesn't know how to import neighbouring files.

If I put `worker.py` at the top level, I need to import from `src`, which defeats the point of the src-layout.

If I put `worker.py` under `src`, the imports work fine, but none of the top-level files are accessible.

Is there a way to make pywrangler build and install the project along with its dependencies (e.g. when a [build system](https://docs.astral.sh/uv/concepts/projects/config/#build-systems) is defined)?
What would be the best place to put `worker.py` in this case?

lemonyte:
Yes, and others at the same level. Although access to top-level files isn't as important as installing the project, so maybe it's best to leave it for another issue.

Here's another example of code that fails when the project is not installed:

```py
# src/package_name/__init__.py
from importlib.metadata import version
__version__ = version("package_name")
print(f"{__version__=}")
```

Raises:
```
importlib.metadata.PackageNotFoundError: No package metadata was found for package_name
```

hoodmane:
I don't think we're planning to build the project like that. Probably what you ought to do is build the wheel yourself and then add the wheel path as a dependency in pyproject.toml for the application. Note also that if we did that the resulting wheel would still not contain the top level files like `pyproject.toml`...

hoodmane:
On the other hand, I agree it is a confusing limitation that you can't include the top level files in the worker.

lemonyte:
> build the wheel yourself and then add the wheel path as a dependency in pyproject.toml for the application

[...]

PR #81, opened 2026-03-17
fix: allow installing local packages from source

ryanking13:
This allows installing a local package from source.

For now, if someone sets the local package as following

```toml
[project]
name = "package-name"
dependencies = [
    "my_project @ ./my_package"
]
```

fails because we do not allow building packages from source through `--no-build` flag.

While this flag is useful for us to prevent accidentally installing native package, it is annoying that users are not possible to test local package.

This is also **required** for us to test `workers-runtime-sdk` in our unittest, as we need to set something like following.

```
dependencies = [
    "workers-runtime-sdk @ ../runtime-sdk"
]
```

Related: https://github.com/cloudflare/workers-py/issues/58#issuecomment-3888722965


packages/cli/src/pywrangler/local_deps.py
```diff
@@ -0,0 +1,55 @@
...
+    for dep in local_deps:
+        name, extras, path = parse_local_dep(dep)
+
+        logger.info(f"Building wheel for local package '{name}' from {path}")
+        if not path.exists():
+            raise RuntimeError(f"Local package path does not exist: {path}")
+
+        dep_out = output_dir / name
+        dep_out.mkdir(parents=True, exist_ok=True)
+        run_command(["uv", "build", "--wheel", str(path), "--out-dir", str(dep_out)])
```
hoodmane, line 48: Do we have a way to make sure these aren't platformed?

ryanking13: Not really... maybe we can check the output name but I wasn't able to think of a very robust way to do it.

PR #87, opened 2026-03-26
feat(workers-runtime-sdk): implement HTMLRewriter SDK in Python

ryanking13:
Implement [HTMLRewriter](https://developers.cloudflare.com/workers/runtime-apis/html-rewriter/) in Python. Original PR is https://github.com/cloudflare/workerd/pull/5945, but I modified it a bit.

The tricky part is how to clean up the handler proxies after running the transform.

- Happy path: after reading the stream succeeds, clean up the PyProxies, done.
- Unhappy path: user can cancel the stream or there can be an exception inside handling the html file
  - unfortunately, it is unclear when we can clean up the proxies as JS HTML rewriter will internally handle the timing when they should cancel the stream operation.
  - in my original PR, I created another Proxy in `_pyodide_entrypoint_helper` that forwards to identity function when the proxy is already distroyed. However, I wasn't sure if it is stable enough.
  - So for now, I just let the proxy leak when exception or cancellation happens.
  - If we have hybrid worker support later, then we should be able to handle it more properly.

While it is not 100% satisfactory, I think it is better than having no SDK as our example was already leaking proxies. WDYT?

### DIFF
--- a/packages/runtime-sdk/src/workers/htmlrewriter.py
+++ b/packages/runtime-sdk/src/workers/htmlrewriter.py
@@ -99,8 +99,7 @@ class HTMLRewriter:
                 # clean up these proxies.
 
         async def cancel(reason):
-            if reader:
-                await reader.cancel(reason)
+            await reader.cancel(reason)
 
         start_proxy = create_proxy(start)
         cancel_proxy = create_proxy(cancel)
--- a/packages/runtime-sdk/src/workers/htmlrewriter.py
+++ b/packages/runtime-sdk/src/workers/htmlrewriter.py
@@ -1,0 +1,122 @@
+import js
+from js import Object
+from pyodide.ffi import JsProxy, create_proxy, to_js
+
+from ._workers import Response, _jsnull_to_none
+
+_ELEMENT_HANDLER_METHODS = ("element", "comments", "text")
+_DOCUMENT_HANDLER_METHODS = ("doctype", "comments", "text", "end")
+
+
+def _create_js_handler(
+    handler: object, method_names: tuple[str, ...], handler_proxies: list[JsProxy]
+) -> JsProxy:
+    """
+    Instead of creating a single proxy for the entire handler object,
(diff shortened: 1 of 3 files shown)


packages/runtime-sdk/src/workers/htmlrewriter.py
```diff
@@ -0,0 +1,121 @@
...
+                if completed:
+                    # Happy path: HTMLRewriter finished processing,
+                    # safe to destroy all proxies
+                    _destroy_proxies()
+                # TODO(later):
+                # On cancel/exception the HTMLRewriter may still hold
+                # internal references to handler method proxies, so we
+                # intentionally leak them to avoid runtime errors.
+                # We should investigate if there is a way to properly
+                # clean up these proxies.
```
dom96, line 99: Why does this happen? What runtime errors do you see if you don't leak them? Could we make changes to HTMLRewriter itself to fix it?

dom96 (commented):
In general looks good to me. I think it's worth investigating if we can make changes to HTMLRewriter itself to handle the cancel/exception case better.

I'd also like @hoodmane to review to make sure the proxy handling is done correctly as I'm not an expert on this.

packages/runtime-sdk/src/workers/htmlrewriter.py
ryanking13: > Why does this happen?

I am not 100% sure, but I think what happens is when the stream is canceled, the HTMLRewriter still does some clean-up job which accesses the handler proxies.

> What runtime errors do you see if you don't leak them?

`Object has already been destroyed`. It is a Pyodide error message that is showed when we access PyProxy that is already destroyed.

> Could we make changes to HTMLRewriter itself to fix it?

I think we can fix it by HTMLRewriter destroying the proxies themselves when they handle cancellation or exception handling, something like the following pseudo code:

```
if (cancelled() || aborted()) {
  for (auto handler: handlers) { if isPyroxy(handler) handler.destroy() }
}
```

PR #88, opened 2026-03-30
fix(workers-py): do not include stale packages in the bundle

ryanking13:
This fixes a bug in `pywrangler sync` that stale packages from the pyproject.toml remains in the `python_module` directory.

For example, the following scenario has a bug:

1) a packages had runtimes dependencies `[a, b]`
2) run `pywrangler sync`
3) removes `b` from the depdency
4) run `pywrangler sync` again
5) (BUG) b still remains in the `python_modules` directory

---

Actually, we could try`uv pip sync` instead of `uv pip install` to deal with this more properly. But requires some more investigation in how we install packages to native/vendor directories, and this is a simpler workaround

### DIFF
--- a/packages/cli/src/pywrangler/sync.py
+++ b/packages/cli/src/pywrangler/sync.py
@@ -188,6 +188,18 @@ def _install_requirements_to_vendor(requirements: list[str]) -> str | None:
         f"Installing packages into [bold]{relative_vendor_path}[/bold]...",
         extra={"markup": True},
     )
+
+    # Clear pyodide venv site-packages so stale packages from previous syncs
+    # don't carry over into python_modules.
+    pyv = get_python_version()
+    site_packages_path = (
+        f"lib/python{pyv}/site-packages" if os.name != "nt" else "Lib/site-packages"
+    )
+    pyodide_site_packages = get_pyodide_venv_path() / site_packages_path
+    if pyodide_site_packages.is_dir():
+        shutil.rmtree(pyodide_site_packages)
+        pyodide_site_packages.mkdir()
+
     with temp_requirements_file(requirements) as requirements_file:
         result = run_command(
             [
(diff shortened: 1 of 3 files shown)


packages/cli/src/pywrangler/sync.py
```diff
@@ -188,6 +188,18 @@ def _install_requirements_to_vendor(requirements: list[str]) -> str | None:
...
+
+    # Clear pyodide venv site-packages so stale packages from previous syncs
+    # don't carry over into python_modules.
+    pyv = get_python_version()
+    site_packages_path = (
+        f"lib/python{pyv}/site-packages" if os.name != "nt" else "Lib/site-packages"
+    )
+    pyodide_site_packages = get_pyodide_venv_path() / site_packages_path
+    if pyodide_site_packages.is_dir():
+        shutil.rmtree(pyodide_site_packages)
```
dom96, line 200: I wonder if this kind of thing will bite some of our users. I can totally see people modifying the files in the python_modules directory while testing and then having their changes deleted due to code like this.

I think we have other code that does similar, so adding this isn't a big deal. But worth thinking about if we want to implement something to prevent accidental deletion.

dom96 (approved):
Thank you for fixing this!

packages/cli/src/pywrangler/sync.py
ryanking13: Yeah, actually I am doing that (modifying python_modules directory) quite often for debugging. Maybe we can add an option to disable `sync` running in the background when calling `pywrangler dev` or `pywrangler deploy`.

Anyway, this change does not hard-block those users as `sync` is not triggered when a user modifies python_modules directory only, it is only triggered when pyproject.toml or wrangler.jsonc is modified, so I'll go ahead and merge this for now.

packages/cli/tests/test_types.py
```diff
@@ -69,5 +69,5 @@ def test_types(tmp_path):
         result = run(["uv", "run", "mypy"], capture_output=True, text=True, check=False)
         assert 'Revealed type is "js.Env"' in result.stdout
         assert 'Revealed type is "js.KVNamespace_iface"' in result.stdout
-        assert 'Revealed type is "builtins.str"' in result.stdout
+        assert 'Revealed type is "str"' in result.stdout
```
ryanking13, line 72: @hoodmane Not sure what has changed in ts-to-python or other type generation related tools, but there was a change in this output

PR #99, opened 2026-04-29
Add entropy import context for packages from workerd

hoodmane:
Import it in a pth file. Will take effect once cloudflare/workerd#6687 is merged and the compat flag is turned on.

### DIFF
--- a/packages/runtime-sdk/src/_workers_sdk_entropy_import_context.pth
+++ b/packages/runtime-sdk/src/_workers_sdk_entropy_import_context.pth
@@ -1,0 +1,1 @@
+import _workers_sdk_entropy_import_context
--- a/packages/runtime-sdk/src/_workers_sdk_entropy_import_context.py
+++ b/packages/runtime-sdk/src/_workers_sdk_entropy_import_context.py
@@ -1,0 +1,183 @@
+"""
+Top level entropy patches for packages
+"""
+
+import sys
+from contextlib import contextmanager
+
+from _cloudflare.allow_entropy import (
+    allow_bad_entropy_calls,
+)
+from _cloudflare.import_patch_manager import (
+    block_calls,
+    register_after_snapshot,
+    register_before_first_request,
+    register_create_patch,
+    register_exec_patch,
+)
+
+
+class STATE:
+    imported_rust_package = False
+    numpy_random = None
+
+
+@register_create_patch("tiktoken._tiktoken")
+@register_exec_patch("cryptography.exceptions")
+@register_exec_patch("jiter")
+@contextmanager
+def rust_package_context(module):
+    """Rust packages need one entropy call if they create a rust hash map at
+    init time.
+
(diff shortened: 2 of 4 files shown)


packages/runtime-sdk/src/_workers_sdk_entropy_import_context.pth
```diff
@@ -0,0 +1 @@
+import _workers_sdk_entropy_import_context
```
ryanking13, line 1: Maybe we should mention that using import in pth file is deprecated and will be removed in the future Python versions:

https://peps.python.org/pep-0829/

packages/runtime-sdk/src/_workers_sdk_entropy_import_context.py
```diff
@@ -0,0 +1,183 @@
...
+from _cloudflare.allow_entropy import (
+    allow_bad_entropy_calls,
+)
+from _cloudflare.import_patch_manager import (
+    block_calls,
+    register_after_snapshot,
+    register_before_first_request,
+    register_create_patch,
+    register_exec_patch,
+)
```
ryanking13, line 17: Should all these live under workerd or can we extract these out to workers-py as well?

Extracting these out would mean we would have duplicated codes in workerd and workers-py, but having less dependency on workerd internal always feels better to me.

dom96: Why would we need to duplicate the code in this case?

dom96 (approved):

packages/runtime-sdk/src/_workers_sdk_entropy_import_context.py
hoodmane: Well we can do the same thing as with the rest of the sdk: Freeze what we currently have in workerd forever and only ever change the code in workers-sdk.

packages/runtime-sdk/src/_workers_sdk_entropy_import_context.pth
hoodmane: Well it's not deprecated until Python 3.15 which doesn't even exist yet. But I can add a comment.

hoodmane: There's a clear path to migrate this as part of pep 829.

packages/runtime-sdk/src/_workers_sdk_entropy_import_context.py
ryanking13: > Why would we need to duplicate the code in this case?

Because we are using it in workerd as well to patch builtin packages. If we can extract them to workers-py as well, then freezing the sdk in workerd should be fine.

ryanking13: > Well we can do the same thing as with the rest of the sdk: Freeze what we currently have in workerd forever and only ever change the code in workers-sdk.

Sounds good to me

ryanking13 (approved):

PR #102, opened 2026-05-12
fix(workers-py): Add --native-tls flag to pywrangler commands

hoodmane:
So the suggestion made by the uv errors actually works.

### DIFF
--- a/packages/cli/src/pywrangler/cli.py
+++ b/packages/cli/src/pywrangler/cli.py
@@ -68,8 +68,15 @@ class ProxyToWranglerGroup(click.Group):
             except ValueError:
                 remaining_args = []
 
+            # --native-tls is consumed by pywrangler (passed through to uv during
+            # sync) and stripped before forwarding to wrangler since wrangler does
+            # not understand this flag.
+            native_tls = "--native-tls" in remaining_args
+            if native_tls:
+                remaining_args = [a for a in remaining_args if a != "--native-tls"]
+
             if cmd_name in ["dev", "publish", "deploy", "versions"]:
-                sync(force=False)
+                sync(force=False, native_tls=native_tls)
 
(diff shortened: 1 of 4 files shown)


packages/runtime-sdk/uv.lock
```diff
@@ -210,7 +210,7 @@ wheels = [
 
 [[package]]
 name = "workers-runtime-sdk"
-version = "1.1.2"
+version = "1.1.3"
```
ryanking13, line 213: Probably we should make our release ci job to update this file too.

packages/cli/src/pywrangler/cli.py
```diff
@@ -68,8 +68,15 @@ def get_command(self, ctx: click.Context, cmd_name: str) -> click.Command:
             except ValueError:
                 remaining_args = []
 
+            # --native-tls is consumed by pywrangler (passed through to uv during
+            # sync) and stripped before forwarding to wrangler since wrangler does
+            # not understand this flag.
+            native_tls = "--native-tls" in remaining_args
+            if native_tls:
+                remaining_args = [a for a in remaining_args if a != "--native-tls"]
```
ryanking13, line 76:
```suggestion
                remaining_args = remaining_args.remove("--native-tls")
```

ryanking13 (approved):
I don't think this would work for Cloudflare warp as there is a crypto algorithm problem remaining in uv, but having this option should be helpful anyways.

dom96:
Does it work? I definitely tried it before and it didn't work for me. For reference since I looked it up again, we have a bug ticket for this BUGS-1799.

hoodmane:
No it doesn't work but at least it makes it not error out when you follow the advice and pass the flag...

dom96 (approved):

PR #103, opened 2026-05-13
chore: copy rpc test from workerd

ryanking13:
Copies https://github.com/cloudflare/workerd/tree/main/src/workerd/server/tests/python/python-rpc here so that we can make sure any SDK changes we make in our rpc call does not break


ryanking13:

dom96 (approved):
Looks good. Can you also add some comments explaining where this was copied from? Also add a comment in workerd to make it clear the test has a new home and the original shouldn't be updated any more.

ryanking13:
recheck

PR #105, opened 2026-05-14
chore: add a unittest template for testing bindings

ryanking13:
This adds a test template for testing bindings such as r2, kv, d1.

I'll start adding tests for those bindings as a follow-up. This PR is mostly a preparation with a small smoke test using kv.


dom96 (approved):
Interesting approach. I suppose we need to run wrangler to be able to use the bindings. I wonder if we should freeze the workerd/wrangler version we run these tests against, to ensure our CI is less susceptible to the world breaking it.

ryanking13:
> I suppose we need to run wrangler to be able to use the bindings.

Yes, it seems like we need a wrangler to test bindings.

> I wonder if we should freeze the workerd/wrangler version we run these tests against, to ensure our CI is less susceptible to the world breaking it.

Maybe, but I guess it is also worth to know when some change in wrangler breaks us unless if we are going to freeze the wrangler version in each pywrangler version as well.

PR #107, opened 2026-05-18
feat(pywrangler): honor [tool.uv.sources] and [tool.uv.workspace] via uv export

LuisDuarte1:
When the project's pyproject.toml has either `[tool.uv.sources]` or `[tool.uv.workspace]`, `parse_requirements` now resolves dependencies via `uv export --no-dev --no-editable --no-hashes --no-emit-project` instead of reading `[project.dependencies]` directly. This honors local-path sources and workspace members, so consumers can depend on in-repo Python packages without having to publish them to PyPI or pre-build wheels into a `find-links` directory.

Workspace-relative paths emitted by `uv export` (e.g. `./packages/foo`) are rewritten to absolute paths in the requirements file so the inner `uv pip install -r` resolves them regardless of cwd.

`_install_requirements_to_vendor` drops `--no-build` when any requirement is a local path, so pure-Python workspace members can be built on the fly into a wheel and vendored into `python_modules/`. Non-workspace deps continue to be resolved from the Pyodide wheel index and `--no-build` is preserved in their case.

Behavior is unchanged for projects that don't use sources or workspaces.

### DIFF
--- a/packages/cli/src/pywrangler/sync.py
+++ b/packages/cli/src/pywrangler/sync.py
@@ -2,9 +2,11 @@ import logging
 import os
 import shutil
 import tempfile
+import tomllib
 from collections.abc import Iterator
 from contextlib import contextmanager
 from pathlib import Path
+from typing import Any
 
 import click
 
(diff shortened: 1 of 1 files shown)


LuisDuarte1:

packages/cli/src/pywrangler/sync.py
```diff
@@ -202,8 +278,33 @@ def _install_requirements_to_vendor(requirements: list[str]) -> str | None:
...
+        # If any requirement is a local path (workspace member / [tool.uv.sources]),
+        # we need to allow building it from source (pure-Python wheel build).
+        # The original `--no-build` is too aggressive in that case; use
+        # `--no-build-package` for each *remote* dep instead.
+        local_paths = {
+            Path(r).resolve()
+            for r in requirements
+            if Path(r).is_absolute() and Path(r).exists()
+        }
+        if local_paths:
```
ryanking13, line 290: This is somewhat tricky part. We don't want to allow users from building non-pure python wheels as they will end up having a native (linux, macos) wheel which cannot run in the worker environment.

But I do agree that we need a way to support local packages for development/workspace purpose. I think we should find a way to combine features such as `--no-build`, `--no-source`, `--no-build-package` to make it work properly.

ryanking13: I had a draft PR for this but didn't make it work (https://github.com/cloudflare/workers-py/pull/81)

ryanking13:
Closing as resolved by https://github.com/cloudflare/workers-py/pull/234

PR #108, opened 2026-05-21
feat(cli): add lockfile support to pywrangler sync command

ryanking13:
pywrangler sync will now generate a `pylock.toml` file that locks the installed packages. Rerunning pywrangler sync will now keep the existing installed versions and will not upgrade packages silently.

Also added `--upgrade` flag to pywrangler sync which upgrades the packages in the lockfile.

### DIFF
--- a/packages/cli/src/pywrangler/resolve.py
+++ b/packages/cli/src/pywrangler/resolve.py
@@ -1,5 +1,4 @@
 import logging
-import tempfile
 import tomllib
 from dataclasses import dataclass, field
 from pathlib import Path
@@ -11,6 +10,7 @@ from .utils import (
     get_uv_pyodide_interp_name,
     read_pyproject_toml,
     run_command,
+    temp_requirements_file,
 )
 
 logger = logging.getLogger(__name__)
@@ -45,12 +45,7 @@ def _compile_requirements(
     exists, ``uv pip compile`` uses it as a constraint source so pinned versions
     are preserved across re-runs (no silent upgrades).
     """
-    with tempfile.NamedTemporaryFile(mode="w", suffix=".in", delete=False) as req_file:
-        req_file.write("\n".join(requirements))
-        req_file.flush()
-        req_in_path = req_file.name
-
-    try:
+    with temp_requirements_file(requirements) as req_in_path:
         cmd = [
             "uv",
             "pip",
@@ -71,8 +66,6 @@ def _compile_requirements(
             cmd.append("--upgrade")
 
         run_command(cmd, cwd=get_project_root(), capture_output=True)
-    finally:
-        Path(req_in_path).unlink(missing_ok=True)
 
     return _read_lockfile_requirements(lockfile_path)
 
--- a/packages/cli/src/pywrangler/resolve.py
+++ b/packages/cli/src/pywrangler/resolve.py
@@ -1,0 +1,139 @@
+import logging
+import subprocess as sp
+import tempfile
+import tomllib
+from dataclasses import dataclass, field
+from pathlib import Path
+
+import click
+
+from .utils import (
+    get_project_root,
+    get_pyodide_index,
+    get_uv_pyodide_interp_name,
+    read_pyproject_toml,
+)
+
+logger = logging.getLogger(__name__)
+
+MANAGED_SDK_PACKAGE = "workers-runtime-sdk"
+LOCKFILE_NAME = "pylock.toml"
+
+
+@dataclass
+class InstallPlan:
+    """Requirements resolved for installation (requirements.txt-style strings)."""
+
+    requirements: list[str] = field(default_factory=list)
+    lockfile: Path | None = None
+
+
+def parse_requirements() -> list[str]:
+    pyproject_data = read_pyproject_toml()
+
+    # Extract dependencies from [project.dependencies]
+    return pyproject_data.get("project", {}).get("dependencies", [])
+
+
+def get_lockfile_path() -> Path:
+    return get_project_root() / LOCKFILE_NAME
+
+
+def _compile_requirements(
+    requirements: list[str],
+    lockfile_path: Path,
+    *,
+    upgrade: bool = False,
+) -> list[str]:
+    """Run ``uv pip compile`` targeting Pyodide and return pinned requirement strings.
+
+    Writes the compiled output to *lockfile_path*. When *lockfile_path* already
+    exists, ``uv pip compile`` uses it as a constraint source so pinned versions
+    are preserved across re-runs (no silent upgrades).
+
+    Uses subprocess directly (instead of run_command) to keep stdout and stderr
+    separate — uv writes the compiled output to stdout but warnings to stderr.
+    """
+    with tempfile.NamedTemporaryFile(mode="w", suffix=".in", delete=False) as req_file:
+        req_file.write("\n".join(requirements))
+        req_file.flush()
+        req_in_path = req_file.name
+
+    try:
+        cmd = [
+            "uv",
+            "pip",
+            "compile",
+            req_in_path,
+            "--python",
+            get_uv_pyodide_interp_name(),
+            "--extra-index-url",
+            get_pyodide_index(),
+            "--index-strategy",
+            "unsafe-best-match",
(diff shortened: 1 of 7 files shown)


packages/cli/src/pywrangler/resolve.py
```diff
@@ -0,0 +1,139 @@
...
+
+def parse_requirements() -> list[str]:
+    pyproject_data = read_pyproject_toml()
+
+    # Extract dependencies from [project.dependencies]
+    return pyproject_data.get("project", {}).get("dependencies", [])
+
+
+def get_lockfile_path() -> Path:
+    return get_project_root() / LOCKFILE_NAME
```
hoodmane: Maybe in utils?

packages/cli/src/pywrangler/resolve.py
```diff
@@ -0,0 +1,139 @@
...
+            "--no-header",
+            "-o",
+            str(lockfile_path),
+        ]
+        if upgrade:
+            cmd.append("--upgrade")
+
+        # TODO: use run_command function instead of sp.run
+        #       like other functions. It requires
+        #       updating the run_command function to handle stdout and stderr separately
```
hoodmane: Looks to me like it'd work fine to use `run_command`.

packages/cli/src/pywrangler/resolve.py
```diff
@@ -0,0 +1,139 @@
...
+    exists, ``uv pip compile`` uses it as a constraint source so pinned versions
+    are preserved across re-runs (no silent upgrades).
+
+    Uses subprocess directly (instead of run_command) to keep stdout and stderr
+    separate — uv writes the compiled output to stdout but warnings to stderr.
+    """
+    with tempfile.NamedTemporaryFile(mode="w", suffix=".in", delete=False) as req_file:
+        req_file.write("\n".join(requirements))
+        req_file.flush()
+        req_in_path = req_file.name
```
hoodmane: Make this into a context manager:
```
def req_file(requirements):
    with tempfile.NamedTemporaryFile(mode="w", suffix=".in") as file:
        file.write("\n".join(requirements))
        file.flush()
        yield file.name
```
Also, don't we already have this logic in other places? Probably this context manager already exists in this code base.

packages/cli/src/pywrangler/resolve.py
```diff
@@ -0,0 +1,139 @@
...
+            encoding="utf-8",
+            check=False,
+        )
+    finally:
+        Path(req_in_path).unlink(missing_ok=True)
+
+    if result.returncode != 0:
+        error_output = (result.stderr or result.stdout or "").strip()
+        logger.error(f"uv pip compile failed:\n{error_output}")
+        raise click.exceptions.Exit(code=1)
```
hoodmane: Should we keep the exit code? For instance, there are a lot of tools that treat keyboard interrupt (130) distinct from other exit codes.

packages/cli/src/pywrangler/sync.py
```diff
@@ -167,16 +158,19 @@ def temp_requirements_file(requirements: list[str]) -> Iterator[str]:
         yield temp_file.name
 
 
-def _install_requirements_to_vendor(requirements: list[str]) -> str | None:
+def _install_requirements_to_vendor(plan: InstallPlan) -> str | None:
     """Install packages to the Pyodide vendor directory.
 
+    When the plan has a lockfile (pylock.toml), installs directly from it.
+    Otherwise falls back to a temp requirements file with dynamic resolution.
```
hoodmane: This is pretty messy. So we have this new `InstallPlan` dataclass but we pretty much only expect to use the requirements file or `pylock.toml`? Are we really modelling an enum here? When will there be no lock file?

hoodmane: Sure looks to me like lockfile is always defined, so the fallback path never happens. Or is the issue that the lock file doesn't always exist? When does it not exist?

packages/cli/src/pywrangler/resolve.py
```diff
@@ -0,0 +1,139 @@
...
+        data = tomllib.load(f)
+
+    results = []
+    for pkg in data.get("packages", []):
+        name = pkg.get("name")
+        version = pkg.get("version")
+        if not name or not version:
+            logger.warning("Skipping malformed lockfile entry: %s", pkg)
+            continue
+        results.append(f"{name}=={version}")
```
hoodmane: A list of `a==b` strings is a bit of a weird way to store the results. I'd kind of think to keep them as a list of `(name, version)` pairs or turn it all the way into a string?

packages/cli/src/pywrangler/sync.py
ryanking13: The reason why I introduced InstallPlan in this PR is to extend it to support workspaces in a follow-up (#107), but maybe it can be done in a separate PR.

packages/cli/src/pywrangler/resolve.py
ryanking13: Right. I originally used the stdout directly but I changed it to use file so this is not needed.

packages/cli/src/pywrangler/sync.py
ryanking13: > Sure looks to me like lockfile is always defined

Correct, let me revise it.

packages/cli/src/pywrangler/resolve.py
ryanking13: Updated to use temp_requirements_file helper method

packages/cli/src/pywrangler/sync.py
ryanking13: So I kept the `InstallPlan` class, but revised it a bit to always use the lockfile, and also moved some helper functions as a method of this class.

packages/cli/src/pywrangler/resolve.py
ryanking13: Sounds good. Updated to tuple

packages/cli/src/pywrangler/sync.py
```diff
@@ -200,29 +179,27 @@ def _install_requirements_to_vendor(requirements: list[str]) -> str | None:
...
+    result = run_command(
+        [
+            "uv",
+            "pip",
+            "install",
+            "--no-build",
+            "-r",
+            str(plan.lockfile),
+            "--preview-features",
+            "pylock",
```
dom96, line 191: Which version was this feature introduced in? Do we need to update our minimum supported uv version? https://github.com/cloudflare/workers-py/blob/main/packages/cli/src/pywrangler/utils.py#L227

packages/cli/tests/test_cli.py
```diff
@@ -365,10 +427,26 @@ def test_sync_command_with_unchanged_timestamps(
...
+    caplog,
 ):
     """Test that the sync command runs when timestamps indicate changes."""
+    from pywrangler.resolve import InstallPlan
+
+    lockfile = test_dir / "pylock.toml"
+    lockfile.write_text(
+        'lock-version = "1.0"\n[[packages]]\nname = "click"\nversion = "8.1.7"\n'
+    )
+    mock_resolve.return_value = InstallPlan(lockfile)
```
dom96: Why do we need this in the test now? Seems like a couple different tests now need a lock file to be present to work. Is this because the lock file is now mandatory? Should we make it optional?

packages/cli/src/pywrangler/sync.py
ryanking13: It was added in 0.6.15 (https://github.com/astral-sh/uv/blob/6feeacc622b8d1a644397fe7d2b8af6d5c3dafcf/changelogs/0.6.x.md?plain=1#L702), so the current minimum version should be fine. But probably we need to update it later anyways to support pyemscripten wheels.

packages/cli/tests/test_cli.py
ryanking13: Changes in this file are unnecessary, thanks for the catch.

Changes in `test_version_sync.py` is necessary as I changed the format of the requirements. The lockfile will be generated if there is no lockfile exists.

dom96 (approved):

PR #110, opened 2026-05-26
fix(runtime-sdk): include missing asgi.py in the wheel

### LINKED ISSUE #109: asgi module not importable in local development after v1.1.3
Summary

  After workers-runtime-sdk v1.1.3 (commit 6e574ca, PR #99), the asgi.py module was removed from the wheel's include list. Since asgi.py is not included in the
  pip-distributed package, import asgi fails in local Python environments, making it impossible to develop ASGI-based applications (FastAPI, Starlette, etc.) locally.

  Background

  In v1.1.2 and earlier, the build configuration was:

  [tool.hatch.build.targets.wheel]
  packages = ["src/workers", "src/asgi.py"]

  PR #99 replaced this with:

  [tool.hatch.build.targets.wheel]
  include = ["src/_*.py", "src/_*.pyi", "src/_*.pth", "src/workers"]

  The asgi.py file no longer matches any of the include patterns and is excluded from the built wheel.

  Why this matters

  While asgi.py is vendored in workerd (src/pyodide/internal/workers-api/src/asgi.py) and available at runtime, local development tooling has no access to it:

  - Type checking / IDE completion: import asgi fails, so type checkers (mypy, pyright) can't resolve asgi.fetch() signatures
  - Local testing: Can't write unit tests that exercise the ASGI adapter locally
  - Static analysis: Tools like pylint/ruff report import asgi as an unresolved import

  The workaround shown in the FastAPI sample (putting import asgi inside the fetch method body) avoids the top-level import error but doesn't solve IDE/type-checker
  issues.

  What I understand so far

  - PR #93 ("Quick fix to top level asgi import") previously ensured asgi.py works at top level by deferring from workers import wait_until
  - PR #99 intentionally removed asgi.py from the wheel build because workerd already vendors a copy
  - PR #100 added the loader shim to suppress .pth warnings in native Python

  Questions

  1. Was the removal of asgi.py from the wheel intentional, or was it an oversight during the build configuration refactoring?
  2. If intentional, what is the recommended approach for local development with ASGI-based apps? Is there a planned replacement (e.g., workers.asgi submodule)?
  3. If it was an oversight, would a PR adding "src/asgi.py" back to the include list be welcome?

  Environment

  - OS: WSL2 on Windows, Ubuntu 24.04.4 LTS (kernel 6.3.0)
  - workers-runtime-sdk >= 1.1.3
  - Python 3.12 / 3.13
  - Local development (not inside workerd runtime)

  ---
  This issue was organized and translated by Claude Code.

ryanking13:
fix #109

### DIFF
--- a/packages/runtime-sdk/pyproject.toml
+++ b/packages/runtime-sdk/pyproject.toml
@@ -68,7 +68,7 @@ warn_unreachable = true
 no_implicit_optional = true
 
 [tool.hatch.build.targets.wheel]
-include = ["src/_*.py", "src/_*.pyi", "src/_*.pth", "src/workers"]
+include = ["src/*.py", "src/_*.pyi", "src/_*.pth", "src/workers"]
 
 [tool.hatch.build.targets.wheel.sources]
 "src" = ""


dom96 (approved):
Thanks!

PR #112, opened 2026-05-26
feat(runtime-sdk): revise type conversion to support bindings more natively

ryanking13:
Note: this includes some breaking changes so it needs a careful look. Also this probaby needs to be merged after #108.

I was experimenting with our current type conversion logic in our SDK to see if a small change is sufficient to make the bindings work more natively. With these changes, I was able to make all the general use cases of storage bindings (R2, KV, D1) work without exposing any Pyodide/JS-specific APIs.

### Changes included

1. Python dict is now converted to JS object by default when passed through the RPC boundary.

This allows python dict to be passed to the RPC function without using `to_js`. For example,

```python
# before
kv.put(key, value, to_js({"metadata": "blah-blah"})
# after
kv.put(key, value, {"metadata": "blah-blah"})
```

This change was actually landed in Pyodide 0.29.0, but was not applied in Python workers.

2. JavaScript `Null` is now converted to Python `None` when coming from RPC boundary.

So that users do not need to care about jsnull, None difference.

```python
# before
kv.get("nonexisting") == pyodide.ffi.jsnull
# after
kv.get(nonexisting") == None
```

3. JavaScript Object coming from RPC boundary is now converted to Python-dict like object, but their attributes are accessible through `["key"]` but also by `.key` (optional)

(This change is not necessary, but I think it improves the user experience)

This allows mixing `.key`-style access and `["key"]`-style access. This helps accessing attributes when the RPC returns complicated, nested object with JsProxy and python dict mixed.

For example, let's say there is a function that returns the following JS object.

```js
{
  key: string,
  data: Blob,
}
```

When it passes through the RPC bounday, it will be converted to a Python dict that is

```python
{
  "key": str,
  "data": JsProxy,
}
```

where Blob is a JsProxy or a [Python class that wrap the JsProxy](https://github.com/cloudflare/workers-py/blob/92e45de388e76566fb5520d38ea5b594b4c79c69/packages/runtime-sdk/src/workers/_workers.py#L617).

In this case, we can now access the `Blob` attribute with

```python
obj = rpc_call()
obj.data.size
```

while previously, users needed to 

```python
obj["data"].size
```

4. Wrapper functions that needs extra type conversion (None => jsnull)

`D1.bind` function accepts `null` as a parameter. Passing `undefined` instead of `null` is a runtime error. To handle this conversion automatically, I made a wrapper function for `.bind` to handle this conversion in the background. This means when there are other bindings that needs to handle the `null` value, we need to create a wrapper function likewise.

Alternatively, we can export `jsnull` type to the users `e.g. from workers import jsnull` and let them pass it to the function, but it makes the user experience less ergonomic.

---

### Changes not included

1. JS function names / parameter names are used as-is.

All the JS function names / parameter names are used as-is, as a camelCase. We do not convert it to Pythonic names.


packages/runtime-sdk/src/workers/_workers.py
```diff
@@ -1003,37 +1043,48 @@ def python_to_rpc(value) -> JsProxy:
...
 
-class _FetcherWrapper:
+class _RPCWrapper:
     def __init__(self, binding):
         self._binding = binding
 
+    def _convert_result(self, result):
+        converted = python_from_rpc(result)
+        if isinstance(converted, JsProxy):
+            return self.__class__(converted)
```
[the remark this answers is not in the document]
ryanking13: I added a comment. We can find a better way to deal with this when we have other bindings that does not return itself.

packages/runtime-sdk/src/workers/_workers.py
```diff
@@ -933,6 +936,34 @@ def _python_from_rpc_default_converter(value, convert, cache):
...
+
+    This is used to convert JS objects to Python dictionaries while maintaining
+    the ability to access keys as attributes.
+    """
+
+    def __getattr__(self, name):
+        try:
+            return self[name]
+        except KeyError:
+            raise AttributeError(name) from None
```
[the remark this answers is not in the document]
ryanking13: Added `__setattr__` and also added a comment about the limitation.

dom96:
> 1. Python dict is now converted to JS object by default when passed through the RPC boundary.
>
> ```
> # before
> kv.put(key, value, to_js({"metadata": "blah-blah"})
> # after
> kv.put(key, value, {"metadata": "blah-blah"})
> ```

This is great, though the ideal would be to make the following work:

```
kv.put(key, value, metadata="blah-blah")
```

I think this would be the most Pythonic option. Though I understand why it might be challenging. I think it's worth discussing.

> 2. JavaScript Null is now converted to Python None when coming from RPC boundary.
> 3. JavaScript Object coming from RPC boundary is now converted to Python-dict like object, but their attributes are accessible through ["key"] but also by .key (optional)

This is great, and I think it makes perfect sense. Is this something we can upstream to Pyodide as well?

> 4. Wrapper functions that needs extra type conversion (None => jsnull)

What is surprising to me is actually that Pyodide converts `None` to `undefined`. Why is that the default? Shouldn't the `undefined` be the special case?

Is there a path for us to change this? Perhaps with some feature flag in Pyodide so that existing Pyodide users aren't broken by this?

ryanking13:
> kv.put(key, value, metadata="blah-blah")

Acutally, this already works. Pyodide converts keyword parameters to JS object as a last parameter, which means the following two are equivalent.

```
kv.put(key, value, { "metadata": "blah-blah" })
kv.put(key, value, metadata="blah-blah")
```

I added some new tests to make sure if works (See: e202e8d702511082c059705a2bfc9f932e26348a)

I am not sure which one we should document though. Maybe we can discuss this to make sure we have all the examples in our documentation to have the same format

> JavaScript Null is now converted to Python None when coming from RPC boundary.
> JavaScript Object coming from RPC boundary is now converted to Python-dict like object, but their attributes are accessible through ["key"] but also by .key (optional)
> This is great, and I think it makes perfect sense. Is this something we can upstream to Pyodide as well?

Yeah, we already have something similar:

- For null to None, it hasn't been long since Pyodide introduced `jsnull` type, and we still have an opt-in flag that converts JS Null to Python None.

- For Python dict working like a JS object, we have `.as_py_json` method in JSProxy that makes the key in the JSProxy accessible through ["key"]

However, the problem is that the existing Python workers that run Pyodide 0.28.2 would still have jsnull so we need to keep this conversion logic unless we drop supporting old Python workers in our sdk.

> What is surprising to me is actually that Pyodide converts None to undefined. Why is that the default? Shouldn't the undefined be the special case?

Well, I think it makes sense to convert `None` to `undefined`.

dom96:
> Well, I think it makes sense to convert None to undefined.

Why?

ryanking13:
> > Well, I think it makes sense to convert None to undefined.
> 
> Why?

Both `None` and `undefined` are used as a default, unprovided value. 


Calling a JS function

```js
function foo(required, optional1, optional2)

foo(1)
```

will set optional values to `undefined` when not provided. Consider if we are wrapping this JS function in Python, such as

```py
def pyFoo(required, optional1=None, optional2=None):
    js.foo(requires, optional1, optional2)

pyFoo(1)
```

It would make more sense to convert None to undefined in this case.

Similarly, a function without a return value will return `undefined` and `None` in JS/Python world respectively.

So overall, I think the None more corresponds to undefined than null.

dom96:
> It would make more sense to convert None to undefined in this case.

I'm not sure that it would. The code is explicitly setting those arguments to "None" which strictly maps to `null` in JS.

That said, I did some digging into this and I can see that this isn't trivial. It seems that in the past Pyodide actually converted both null and undefined to `None` in Python as well (and I now see you wrote about that above as well).

I think there is a bunch of trade offs here between ergonomics and safety. It's not immediately clear what the best approach is. One good argument I can see for not having a `jsundefined` is that you would then need to write these kinds of checks all over Python FFI code:

```python
x = js_func()
if x is None or x is jsundefined: ...
```

Though on the other hand, it seems you also need to be aware that:

```python
payload = {
    "name": "Alice",
    "age": None
}

json_string = js.JSON.stringify(to_js(payload, ...))
```

Will end up deleting the `age` key, which could be a gotcha for many.

ryanking13:
Yeah, we don't have a perfect solution for this as it is a fundamental difference between the two languages. Overall, converting None to undefined felt more natural for several years in Pyodide to the JS developers.

But as the solution we are looking for here is not something in general but something that works for CF workers, let me see if converting None to null works for bindings

ryanking13:
Okay, looks like converting None to null works for all the storage bindings we are testing in this PR, I guess Cloudflare bindings handle undefined/null pretty well at least.

hoodmane:
`bindings-test` should ideally use pytest rather than this ad-hoc test runner. Can fix it in a followup.

packages/cli/tests/bindings-test/src/test_d1.py
```diff
@@ -0,0 +1,492 @@
...
+        TEST_TABLE,
+        TEST_TABLE_TYPES,
+        TEST_TABLE_BATCH,
+        EXEC_TABLE,
+        EXEC_MULTI_TABLE,
+    ]:
+        try:
+            await db.exec(f"DROP TABLE IF EXISTS {table}")
+        except Exception:
+            pass
```
hoodmane, line 22: What sorts of exceptions can happen here? Can we delete the blanket catch and see if the test fails?

packages/cli/tests/bindings-test/src/kv_test.py
```diff
@@ -1,17 +1,307 @@
...
+    while True:
+        options = {"prefix": "_test:", "limit": 1000}
+        if cursor:
+            options["cursor"] = cursor
+        result = await kv.list(options)
+        for key_entry in result["keys"]:
+            await kv.delete(key_entry["name"])
+        if result["list_complete"]:
+            break
+        cursor = result.get("cursor")
```
hoodmane: Maybe:
```suggestion
    cursor = None
    options = {"prefix": "_test:", "limit": 1000}
    while True:        
        result = await kv.list(options)
        for key_entry in result["keys"]:
            await kv.delete(key_entry["name"])
        if result["list_complete"]:
            break
        options["cursor"] = result.get("cursor")
```

packages/cli/tests/bindings-test/src/kv_test.py
```diff
@@ -1,17 +1,307 @@
...
+
+
+async def test_put_and_get_text(env):
+    kv = env.KV
+    await _cleanup_kv(kv)
+    key = "_test:put_get_text"
+    value = "hello from KV"
+    await kv.put(key, value)
+    result = await kv.get(key)
+    assert result == value, f"expected {value!r}, got {result!r}"
```
hoodmane: We should really just switch this to using pytest and then we'll get nice errors automatically.

hoodmane:
I opened #119 which makes the bindings tests use pytest.

packages/cli/tests/bindings-test/src/test_d1.py
ryanking13: It is a JsException with "invalid syntax..." I updated it to use pytest.raises with match statement

packages/cli/tests/bindings-test/src/kv_test.py
ryanking13: sounds good.

packages/cli/tests/bindings-test/src/kv_test.py
ryanking13: Yeah, it seems like ai agents really like to add those messages. I think sometimes helps, but most of them are redundant. I cleared team up.

packages/cli/tests/bindings-test/src/test_kv.py
```diff
@@ -0,0 +1,300 @@
...
+    assert result["value"] is None
+    assert result["metadata"] is None
+
+
+@pytest.mark.asyncio
+async def test_put_with_expiration_ttl(env):
+    kv = env.KV
+    await _cleanup_kv(kv)
+    key = "_test:expiration_ttl"
+    await kv.put(key, "expires soon", {"expirationTtl": 60})
```
dom96: Some tests using this syntax would be good too. Might be best to use this as the default tbh, since it's more Pythonic and is what we should recommend to folks (even though these are tests, some might use them as examples)

```suggestion
    await kv.put(key, "expires soon", expirationTtl=60)
```

packages/cli/tests/workerd-test/python-rpc/worker.js
```diff
@@ -34,19 +34,32 @@ export default {
       1,
       'test',
       [1, 2, 3],
-      new Map([['key', 42]]),
```
dom96, line 37: We should make sure we document this change somewhere. Should we start a changelog in this repo for things like this?

packages/runtime-sdk/src/workers/_workers.py
```diff
@@ -933,6 +936,41 @@ def _python_from_rpc_default_converter(value, convert, cache):
...
+
+    def __setattr__(self, name, value):
+        self[name] = value
+
+
+def _normalize_result(obj):
```
dom96: Just to be consistent with `_replace_none_with_jsnull`

```suggestion
def _replace_jsnull_with_none(obj):
```

packages/runtime-sdk/src/workers/_workers.py
```diff
@@ -1003,37 +1060,58 @@ def python_to_rpc(value) -> JsProxy:
...
         default_converter=_python_to_rpc_default_converter,
-        dict_converter=js.Map.new,
+        dict_converter=Object.fromEntries,
     )
 
     return result
 
 
-class _FetcherWrapper:
+class _RPCWrapper:
```
dom96: This should be called `_BindingWrapper` IMO

packages/cli/tests/bindings-test/src/test_kv.py
ryanking13: Yeah, I have added tests that have a name with `kwargs_*` to make sure both `{"expirationTtl": 60}` and `expirationTtl=60` work.

I am still not sure if we want to use it as a default, but if both you and Hood thinks that it is more Pythonic, then yeah, let me change it as a default.

packages/cli/tests/workerd-test/python-rpc/worker.js
ryanking13: Looks like the changelog is auto generated in this repository, so I tried to add a long, descriptive commit message that explains the change, but let me double check.

packages/runtime-sdk/src/workers/_workers.py
ryanking13: Sounds good

packages/cli/tests/bindings-test/src/test_kv.py
ryanking13: Done

ryanking13:
Merged https://github.com/cloudflare/workers-py/pull/113 into this branch as well, so that all the type-conversion changes can land in a single release.

packages/cli/tests/bindings-test/src/worker_durable_object.py
```diff
@@ -0,0 +1,213 @@
...
+
+    async def test_storage_delete_multiple(self):
+        await self.ctx.storage.deleteAll()
+        await self.ctx.storage.put({"d1": 1, "d2": 2, "d3": 3})
+        count = await self.ctx.storage.delete(["d1", "d2"])
+        assert count == 2, f"expected 2, got {count!r}"
+
+    async def test_storage_list(self):
+        await self.ctx.storage.deleteAll()
+        await self.ctx.storage.put({"list:a": 1, "list:b": 2, "list:c": 3, "other": 99})
```
dom96, line 64: Nice that this now works.

dom96 (approved):
Nice, looks great and fixes some of the bugs I found in my agentic loop testing. Thank you!

PR #114, opened 2026-05-29
chore: run workerd tests against multiple compat dates

ryanking13:
Runs workerd tests in multiple compat dates to ensure that we don't break older python workers by mistake.

### DIFF
--- a/packages/cli/tests/conftest.py
+++ b/packages/cli/tests/conftest.py
@@ -2,6 +2,7 @@
 # Each date exercises a different Python version inside the worker runtime:
 #   - "2025-09-01" -> Python 3.12 (before the 2025-09-29 cutover)
 #   - "2026-01-01" -> Python 3.13 (after the 2025-09-29 cutover)
+# TODO: use compat flag instead of compat date
 from pathlib import Path
 
 COMPAT_DATES: list[str] = ["2025-09-01", "2026-01-01"]
--- a/packages/cli/tests/conftest.py
+++ b/packages/cli/tests/conftest.py
@@ -1,0 +1,11 @@
+# Compat dates used to parametrize workerd and bindings integration tests.
+# Each date exercises a different Python version inside the worker runtime:
+#   - "2025-09-01" -> Python 3.12 (before the 2025-09-29 cutover)
+#   - "2026-01-01" -> Python 3.13 (after the 2025-09-29 cutover)
+from pathlib import Path
+
+COMPAT_DATES: list[str] = ["2025-09-01", "2026-01-01"]
+
+
+def replace_compat_date(file: Path, compat_date: str) -> None:
+    file.write_text(file.read_text().replace("%COMPAT_DATE", compat_date))
--- a/packages/cli/tests/test_bindings.py
+++ b/packages/cli/tests/test_bindings.py
@@ -18,6 +18,7 @@ from typing import Any, Literal, NotRequired, TypedDict
 
 import pytest
 import requests
+from conftest import COMPAT_DATES, replace_compat_date
 
 TEST_DIR: Path = Path(__file__).parent
 BINDINGS_TEST_DIR: Path = TEST_DIR / "bindings-test"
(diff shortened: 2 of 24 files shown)


packages/cli/tests/test_in_workerd.py
```diff
@@ -45,17 +47,34 @@ def embed(dir: Path, root: Path, level: int = 0):
...
+    # pywrangler sync fails to install pyodide packages in unittest environment + Python 3.12 + Linux
+    # This is reproducable only in the unittest environment, and doesn't happen
+    # when running the same worker manually.
+    if (
+        test_dir.name == "sdk"
+        and compat_date < "2025-09-29"
+        and sys.platform == "linux"
+    ):
+        pytest.xfail("pywrangler sync + uv + pyodide 3.12 on Linux")
+
```
ryanking13, line 69: Something is wrong here, but I cannot reproduce this in non-unittest environment. Running the same worker with pywrangler manually works fine.

packages/cli/tests/conftest.py
```diff
@@ -0,0 +1,11 @@
+# Compat dates used to parametrize workerd and bindings integration tests.
+# Each date exercises a different Python version inside the worker runtime:
+#   - "2025-09-01" -> Python 3.12 (before the 2025-09-29 cutover)
+#   - "2026-01-01" -> Python 3.13 (after the 2025-09-29 cutover)
+from pathlib import Path
+
+COMPAT_DATES: list[str] = ["2025-09-01", "2026-01-01"]
```
dom96, line 8: Could we use compat flag names here instead of compat dates?

packages/cli/tests/test_bindings.py
```diff
@@ -113,15 +123,16 @@ def dev_server(tmp_path_factory: pytest.TempPathFactory) -> Generator[str]:
         process.wait()
 
 
-_test_suite_cache: dict[str, SuiteResults] = {}
+_test_suite_cache: dict[tuple[str, str], SuiteResults] = {}
```
dom96, line 127: document what the fields in this tuple represent, or better yet use class with field names

packages/cli/tests/test_in_workerd.py
dom96: What's the failure?

dom96 (approved):
Mostly looks good, but I left a few suggestions.

packages/cli/tests/test_in_workerd.py
ryanking13:
```
service python-sdk: Uncaught Error: PythonError: Traceback (most recent call last):
  File "/lib/python312.zip/_pyodide/_base.py", line 502, in eval_code
    .run(globals, locals)
     ^^^^^^^^^^^^^^^^^^^^
  File "/lib/python312.zip/_pyodide/_base.py", line 340, in run
    coroutine = eval(self.code, globals, locals)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<exec>", line 9, in <module>
  File "/lib/python312.zip/_pyodide/_base.py", line 629, in pyimport_impl
    res = __import__(stem, fromlist=fromlist)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/session/metadata/worker.py", line 15, in <module>
    import pytest
ModuleNotFoundError: No module named 'pytest'
```

It looks like uv is installing the package in the wrong place. Probably the path to the virtualenv is messed up for some reason.

packages/cli/tests/conftest.py
ryanking13: I think it is possible, but I found that mixing `python_workers_20250116`, `no_python_workers_20250116` is a bit confusing. Maybe let's do that when we found issues using compat date.

packages/cli/tests/test_bindings.py
ryanking13: Updated comment.

PR #115, opened 2026-05-30
fix(cli): don't install Pyodide interpreter onto user's PATH

GertBurger:
`uv python install` will by default place a symlink in the user's PATH (`~/.local/bin` in my case) which is quite unexpected late on a friday night :rofl:

This PR removes the `uv python install` line, as it is not required. uv will install the requested python version/variant during venv creation and this doesn't pollute the host system.

claude wrote the test and it seems to function correctly on my system. Not too sure if it is really useful in the long term.

I only tested on linux.


GertBurger:

ryanking13 (approved):
Thanks, sounds okay to me. If it introduces any regression, maybe we can use `--no-bin` flag instead to avoid adding the binary to PATH.

PR #118, opened 2026-06-01
chore: add more tests for service bindings

ryanking13:
This on top of #113 (I want stacked PRs...)

Adds some service-bindings related tests. This PR purely updates test without changing SDK.

Some of them are duplicate of [sdk_tests.py](https://github.com/cloudflare/workers-py/blob/main/packages/cli/tests/workerd-test/sdk/tests/test_sdk.py) but I wanted to locate the bindings related tests in the same directory + run a full e2e test using pywrangler.

### DIFF
--- a/packages/cli/tests/bindings-test/src/service_test.py
+++ b/packages/cli/tests/bindings-test/src/service_test.py
@@ -1,0 +1,170 @@
+async def test_identity_primitives(env):
+    svc = env.SERVICE_BINDING
+    assert await svc.identity("hello") == "hello"
+    assert await svc.identity(42) == 42
+    assert await svc.identity(3.14) - 3.14 < 0.001
+    assert await svc.identity(True) is True
+    assert await svc.identity(False) is False
+    assert await svc.identity(None) is None
+
+
+async def test_identity_dict(env):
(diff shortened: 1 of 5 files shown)


dom96:
GitHub supports stacked PRs now, doesn't it? https://github.github.com/gh-stack/

ryanking13:
Yeah, but I think it is still in the private beta that is not enabled for this org.

packages/cli/tests/bindings-test/src/service_test.py
```diff
@@ -0,0 +1,170 @@
+async def test_identity_primitives(env):
+    svc = env.SERVICE_BINDING
+    assert await svc.identity("hello") == "hello"
+    assert await svc.identity(42) == 42
+    assert await svc.identity(3.14) - 3.14 < 0.001
+    assert await svc.identity(True) is True
+    assert await svc.identity(False) is False
+    assert await svc.identity(None) is None
```
dom96, line 8: How are these different to the tests we already do in python-rpc workerd tests?

dom96 (commented):
If we're duplicating tests, maybe we could deduplicate somehow or remove the other test cases if these give better test coverage?

PR #123, opened 2026-06-15
feat: auto-convert Python objects that are passed to/from Queue Binding

ryanking13:
Wrap [Cloudflare Queue](https://developers.cloudflare.com/queues/) with our bindings wrapper so that users can see the native Python objects not JSProxies. By doing this, we can eliminate `to_js` and `to_py` when using Queue in Python workers.

While Queue is not our top priority, this is to remove `to_js` from our Python workers doc (https://developers.cloudflare.com/workers/languages/python/examples/#publish-to-a-queue)

### DIFF
--- a/packages/cli/tests/bindings-test/src/test_queue.py
+++ b/packages/cli/tests/bindings-test/src/test_queue.py
@@ -16,7 +16,7 @@ async def _send_and_receive(env, body, **send_opts):
 
     RECEIVED_MESSAGES.clear()
     await env.TEST_QUEUE.send(body, **send_opts)
-    await asyncio.sleep(2)
+    await asyncio.sleep(2)  # Wait for the message to be processed
     assert len(RECEIVED_MESSAGES) > 0, "no messages received by consumer"
     return RECEIVED_MESSAGES[-1]
 
--- a/packages/cli/tests/bindings-test/src/test_queue.py
+++ b/packages/cli/tests/bindings-test/src/test_queue.py
@@ -1,0 +1,114 @@
+import asyncio
+import sys
+
+import pytest
+
+# FIXME: This doesn't seem to happen in prod environment.
+#        But for some reason, pytest segfaults after running tests in dev environment (workerd).
+pytestmark = pytest.mark.skipif(
+    sys.version_info < (3, 13),
+    reason="pytest segfaults after running tests",
+)
+
+
+async def _send_and_receive(env, body, **send_opts):
+    from worker import RECEIVED_MESSAGES
+
+    RECEIVED_MESSAGES.clear()
+    await env.TEST_QUEUE.send(body, **send_opts)
+    await asyncio.sleep(2)
+    assert len(RECEIVED_MESSAGES) > 0, "no messages received by consumer"
+    return RECEIVED_MESSAGES[-1]
+
+
+@pytest.mark.asyncio
+async def test_send_string(env):
+    msg = await _send_and_receive(env, "hello queue")
+    assert msg["body"] == "hello queue"
+    assert isinstance(msg["id"], str)
+    assert msg["attempts"] >= 1
+
+
+@pytest.mark.asyncio
+async def test_send_dict(env):
+    msg = await _send_and_receive(env, {"key": "value", "number": 42})
+    assert msg["body"]["key"] == "value"
+    assert msg["body"]["number"] == 42
+
+
+@pytest.mark.asyncio
+async def test_send_number(env):
+    msg = await _send_and_receive(env, 123)
+    assert msg["body"] == 123
+
+
+@pytest.mark.asyncio
+async def test_send_with_content_type(env):
+    msg = await _send_and_receive(env, "text message", contentType="text")
+    assert msg["body"] == "text message"
+
+
+@pytest.mark.asyncio
+async def test_send_none(env):
+    msg = await _send_and_receive(env, None)
+    assert msg["body"] is None
+
+
+@pytest.mark.asyncio
+async def test_send_bool(env):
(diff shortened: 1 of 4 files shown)


packages/runtime-sdk/src/workers/_workers.py
```diff
@@ -319,8 +319,13 @@ def _manage_pyproxies():
...
 
-def _is_js_instance(val, js_cls_name):
-    return hasattr(val, "constructor") and val.constructor.name == js_cls_name
+def _is_js_instance(val, js_cls_names: str | set[str]):
+    if not hasattr(val, "constructor"):
+        return False
+    name = val.constructor.name
+    if isinstance(js_cls_names, set):
+        return name in js_cls_names
+    return name == js_cls_names
```
dom96, line 328: Perhaps a bit subjective, but I would separate this out like this:

```suggestion
def _get_js_cls_name(val) -> str | None:
    return val.constructor.name if hasattr(val, "constructor") else None

def _is_js_instance(val, js_cls_name: str) -> bool:
    return _get_js_cls_name(val) == js_cls_name

def _is_any_js_instance(val, js_cls_names: set[str]) -> bool:
    return _get_js_cls_name(val) in js_cls_names
```

packages/runtime-sdk/src/workers/_workers.py
```diff
@@ -1114,6 +1119,11 @@ def _convert_result(self, result):
             # TODO: This is a bit of a hack. We should revisit when there are more
             # bindings to support with different patterns.
             return self.__class__(converted)
+        if isinstance(converted, list):
+            return [
+                self.__class__(item) if isinstance(item, JsProxy) else item
```
dom96, line 1126: What does `__class__` of JsProxy actually return here?

A comment explaining what this code is for would be good

dom96 (approved):
A few suggestions, but overall nice!

packages/cli/tests/bindings-test/src/test_queue.py
```diff
@@ -0,0 +1,145 @@
+import asyncio
+import sys
+
+import pytest
+
+pytestmark = [
+    pytest.mark.skipif(
+        sys.version_info < (3, 13),
+        reason="pytest segfaults after running tests",
```
hoodmane: Seems like something we should understand? I guess 3.12 is only best effort for new features but still.

packages/cli/tests/bindings-test/src/test_queue.py
```diff
@@ -0,0 +1,145 @@
...
+    global _send_cache
+    if _send_cache is not None:
+        return _send_cache
+
+    from worker import RECEIVED_MESSAGES
+
+    RECEIVED_MESSAGES.clear()
+
+    # Send everything at once to reduce the overall test time
+    await env.TEST_QUEUE.send("hello queue")
```
hoodmane: Isn't it weird that a function called get_send_results sends a bunch of stuff? I wouldn't expect a function named like that to do IO.

hoodmane: I think the tests would make more sense if we sent just the appropriate things in each test?

packages/runtime-sdk/src/workers/_workers.py
```diff
@@ -1567,3 +1579,19 @@ def __init__(self, ctx: "ExecutionContext", env: "Env"):
...
         _wrap_workflow_step(cls)
+
+
+def _wrap_queue_handler(cls):
+    queue_fn = getattr(cls, "queue", None)
+    if queue_fn is None:
+        return
+
+    @functools.wraps(queue_fn)
+    async def wrapped_queue(self, batch, *_args, **_kwargs):
```
hoodmane: I don't think it's a good idea to accept args and kwargs and just ignore them.

packages/cli/tests/bindings-test/src/test_queue.py
ryanking13: I originally did that (See the change in this commit: https://github.com/cloudflare/workers-py/pull/123/commits/90e92bcdf47b6a32a67b533f361fcd87ca496178)

But I found that sending each message to the queue takes around 2 seconds for the receiver to receive the message in the workerd envrionmnet. So running test sequentially increases the overall tests time a lot.

I'll update the function name and the comment to explain it more clearly.

packages/runtime-sdk/src/workers/_workers.py
ryanking13: Right, that should be forwarded.

packages/cli/tests/bindings-test/src/test_queue.py
ryanking13: Yeah, I'll do some more digging. I am thinking that it is something to do with our webloop implementation.

packages/runtime-sdk/src/workers/_workers.py
ryanking13: It is basically the same as `_BindingWrapper(item)` (or the subclass if it is a child class of it). I'll update the comment

packages/runtime-sdk/src/workers/_workers.py
ryanking13: Thanks for the suggestion. I think it is more pythonic to have a one function that handles multiple input types, so let me keep this as-is for this PR.

packages/cli/tests/bindings-test/src/test_queue.py
ryanking13: The problem was again the `_cancel_all_tasks`. I updated it from no-op to original implementation + handles `InvalidStateError` exception that was problematic in other tests.

PR #126, opened 2026-06-16
fix: ensure Worker subclasses are wrapped only once

### LINKED ISSUE #125: Unable to Access ctx.storage During DurableObject Initialization
I have a `DurableObject`:

```python
from typing import TYPE_CHECKING

from workers import DurableObject

if TYPE_CHECKING:
    from js import DurableObjectState, Env


class MyDurableObject(DurableObject):
    def __init__(self, ctx: "DurableObjectState", env: "Env") -> None:
        super().__init__(ctx, env)
        self._sql = ctx.storage.sql
```

In a clean environment, I run the development server using `uv run pywrangler dev`. When the class constructor is called, I encounter an exception indicating that the object `ctx.storage.constructor` does not have an attribute `name`:

```plain-text
...
✘ [ERROR] Uncaught PythonError: Traceback (most recent call last):

    File "/session/metadata/python_modules/workers/_workers.py", line 1491, in wrapped_init
      original_init(self, *args, **kwargs)
      ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
    File "/session/metadata/python_modules/workers/_workers.py", line 1491, in wrapped_init
      original_init(self, *args, **kwargs)
      ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
    File "/session/metadata/utils.py", line 52, in __init__
      self._sql = ctx.storage.sql
                  ^^^^^^^^^^^
    File "/session/metadata/python_modules/workers/_workers.py", line 1182, in __getattr__
      if _is_js_instance(result, "DurableObjectStorage"):
         ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/session/metadata/python_modules/workers/_workers.py", line 323, in _is_js_instance
      return hasattr(val, "constructor") and val.constructor.name == js_cls_name
                                             ^^^^^^^^^^^^^^^^^^^^
  AttributeError: 'function' object has no attribute 'name'
...
```

wrangler 4.100.0

workers-py==1.14.0
workers-runtime-sdk==1.3.0

I have not been able to identify the true cause of the error. What could help me resolve this issue?

ryanking13:
fixes #125 

Our `_wrap_subclass` had an issue that when there is a subclass that inherits the wrapped WorkerEntrypoint/DurableObject,  the `_wrap_subclass` is called multiple times, double wrapping the `env` or `ctx` attributes.

We always had that issue but it was surfaced by my recent change in the DO (#112)

This fixes it by adding a guarder variable and applying the wrapper function only once.

### DIFF
--- a/packages/runtime-sdk/src/workers/_workers.py
+++ b/packages/runtime-sdk/src/workers/_workers.py
@@ -1475,18 +1475,27 @@ async def _do_call(entrypoint, name, config, callback, *results):
     return result
 
 
+_INIT_WRAPPED = "__workers_init_wrapped__"
+
+
 def _wrap_subclass(cls):
     # Override the class __init__ so that we can wrap the `env` in the constructor.
     original_init = cls.__init__
 
     def wrapped_init(self, *args, **kwargs):
-        args = list(args)
-        if len(args) > 0:
-            _pyodide_entrypoint_helper.patchWaitUntil(args[0])
-            if issubclass(cls, DurableObject):
-                args[0] = DurableObjectContext(args[0])
-        if len(args) > 1:
-            args[1] = _EnvWrapper(args[1])
+        # Guard against double-wrapping in multi-level inheritance.
+        # __init_subclass__ fires for every subclass, so each level installs
+        # its own wrapped_init. The per-instance flag ensures ctx/env are
+        # wrapped only once, by the outermost wrapped_init.
+        if not hasattr(self, _INIT_WRAPPED):
+            args = list(args)
+            if len(args) > 0:
+                _pyodide_entrypoint_helper.patchWaitUntil(args[0])
+                if issubclass(cls, DurableObject):
+                    args[0] = DurableObjectContext(args[0])
+            if len(args) > 1:
+                args[1] = _EnvWrapper(args[1])
+            setattr(self, _INIT_WRAPPED, True)
 
         original_init(self, *args, **kwargs)
 
--- a/packages/runtime-sdk/src/workers/_workers.py
+++ b/packages/runtime-sdk/src/workers/_workers.py
@@ -1475,27 +1475,18 @@ async def _do_call(entrypoint, name, config, callback, *results):
     return result
 
 
-_INIT_WRAPPED = "__workers_init_wrapped__"
-
-
 def _wrap_subclass(cls):
     # Override the class __init__ so that we can wrap the `env` in the constructor.
     original_init = cls.__init__
 
     def wrapped_init(self, *args, **kwargs):
-        # Guard against double-wrapping in multi-level inheritance.
-        # __init_subclass__ fires for every subclass, so each level installs
-        # its own wrapped_init. The per-instance flag ensures ctx/env are
-        # wrapped only once, by the outermost wrapped_init.
-        if not hasattr(self, _INIT_WRAPPED):
-            args = list(args)
-            if len(args) > 0:
-                _pyodide_entrypoint_helper.patchWaitUntil(args[0])
-                if issubclass(cls, DurableObject):
-                    args[0] = DurableObjectContext(args[0])
-            if len(args) > 1:
-                args[1] = _EnvWrapper(args[1])
-            setattr(self, _INIT_WRAPPED, True)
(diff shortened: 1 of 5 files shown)


packages/runtime-sdk/src/workers/_workers.py
```diff
@@ -1475,18 +1475,27 @@ async def _closure():
...
-            _pyodide_entrypoint_helper.patchWaitUntil(args[0])
-            if issubclass(cls, DurableObject):
-                args[0] = DurableObjectContext(args[0])
-        if len(args) > 1:
-            args[1] = _EnvWrapper(args[1])
+        # Guard against double-wrapping in multi-level inheritance.
+        # __init_subclass__ fires for every subclass, so each level installs
+        # its own wrapped_init. The per-instance flag ensures ctx/env are
+        # wrapped only once, by the outermost wrapped_init.
+        if not hasattr(self, _INIT_WRAPPED):
```
dom96: Early returns are nicer

```suggestion
        if hasattr(self, _INIT_WRAPPED): return
        setattr(self, _INIT_WRAPPED, True)
        ...
```

dom96 (approved):
Thanks! See comment for small refactoring nit.

packages/runtime-sdk/src/workers/_workers.py
hoodmane: _INIT_WRAPPED should be a property on the class and not on the instance.

ryanking13: Uh, right. It should not be in the instance level. Actually I noticed that setting it in the class level also can be a problem as there will still be a double-wrap as parent and child class would can both wrapped.

ryanking13: Okay, instead of introducing that attribute, I just updated the `_init_subclass_` functions to only apply the wrapper to the base class.

packages/runtime-sdk/src/workers/_workers.py
```diff
@@ -1549,7 +1550,10 @@ def __init__(self, ctx: "ExecutionContext", env: "Env"):
         self.env = env
 
     def __init_subclass__(cls, **_kwargs: Any):
-        _wrap_subclass(cls)
+        # Make sure we do not apply the wrapper multiple times
+        # when inheriting from the base class
+        if WorkerEntrypoint in cls.__bases__:
```
hoodmane: This is much nicer but I'm a bit confused why `WorkerEntrypoint` is not in `cls.__bases__` the first time.

ryanking13: You mean why `WorkerEntrypoint` doesn't have itself in the `cls.__bases__`? Following is the explanation of `__bases__` attribute:

> A [tuple](https://docs.python.org/3/library/stdtypes.html#tuple) containing the class’s bases. In most cases, for a class defined as class X(A, B, C), X.__bases__ will be exactly equal to (A, B, C).

hoodmane: Sounds kind of fishy to me. So what happens if we do:
```py
class DirectSub(WorkerEntrypoint):
   ...

class SubSub(DirectSub):
    ... 
```
what about:
```py
class DirectSub(WorkerEntrypoint):
   def __init__(...):
       ...

class SubSub(DirectSub):
   def __init__(...):
       super().__init__(...)
```

hoodmane: I guess what you're saying is `SubSub` won't have `WorkerEntrypoint` in bases. But you are allowed to do this as well:
```py
class SubSub(DirectSub, WorkerEntrypoint):
   ...
```

ryanking13: > I guess what you're saying is SubSub won't have WorkerEntrypoint in bases. But you are allowed to do this as well:

Right. Well, I am not sure if people will do that... but let me think if I can prevent that as well.

ryanking13: Updated the check to handle multi inheritance case as well

PR #127, opened 2026-06-16
fix: make iterable returned from rpc call work correctly 

ryanking13:
Fixes another bug reported in #125

Durable Objects Storage returns a cursor when `ctx.storage.sql.exec("...")`  is called. This cursor is an iterable that users can do `for row in cursor`.

However, our `_BindingWrapper` does not handle iterable correctly, hence users were getting the following error message.

```
attribute name must be string, not 'int'
```

### DIFF
--- a/packages/runtime-sdk/src/workers/_workers.py
+++ b/packages/runtime-sdk/src/workers/_workers.py
@@ -1105,6 +1105,13 @@ class _BindingWrapper:
     def __init__(self, binding):
         self._binding = binding
 
+    @property
+    def _real_name(self):
+        try:
+            return self._binding.constructor.name
+        except Exception:
+            return type(self).__name__
+
     def _convert_result(self, result):
         converted = python_from_rpc(result)
         if isinstance(converted, JsProxy):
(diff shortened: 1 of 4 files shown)


packages/runtime-sdk/src/workers/_workers.py
```diff
@@ -1105,6 +1105,13 @@ class _BindingWrapper:
     def __init__(self, binding):
         self._binding = binding
 
+    @property
+    def _real_name(self):
+        try:
+            return self._binding.constructor.name
+        except Exception:
```
dom96: Can you check for the existence of `constructor` and `name` instead of catching all exceptions?

dom96: or maybe give this method a name like `_real_name_for_exception_msg` to make sure its intent is clear

dom96 (approved):

packages/runtime-sdk/src/workers/_workers.py
ryanking13: Updated it to use `_get_js_constructor_name` that I added in #128

PR #128, opened 2026-06-17
fix: fix ReadableStream being incorrectly wrapped by BindingWrapper

ryanking13:
This fixes a bug that the `ReadableStream` object returned from rpc call is incorrectly wrapped with `_BindingWrapper`.

In #112, I wrapped all JsProxies returned from rpc call with `_BindingWrapper`, so that nested rpc objects can be handled properly. However, this was too aggressive that it ended up wrapping objects such as `ReadableStream` with the `_BindingWrapper` as well.

This should not be not a problem normally because `_BindingWrapper` basically pass-through all the attributes and parameters. However, when we are accessing the `constructor` attribute, which is both an object and callable, we wrap it with a Python function and loose the access to its inner attributes, leading to the following error.

```
    File "introspection.py", line 88, in wrapper_func
      return python_to_rpc(await result)
                           ^^^^^^^^^^^^
    File "/session/metadata/entry.py", line 29, in fetch
      return Response(obj.body, headers={"etag": obj.httpEtag})
    File "/session/metadata/python_modules/workers/_workers.py", line 418, in __init__
      if body.constructor.name not in RESPONSE_ACCEPTED_TYPES:
         ^^^^^^^^^^^^^^^^^^^^^
  AttributeError: 'function' object has no attribute 'name'
```

This PR is a workaround, avoid wrapping JS objects that can directly passed to the Response object. In the long term, I think the proper fix is to create a Python wrapper for ReadableStream and handle it inside the `python_from_rpc` function.

### DIFF
--- a/packages/runtime-sdk/src/workers/_workers.py
+++ b/packages/runtime-sdk/src/workers/_workers.py
@@ -392,14 +392,18 @@ RESPONSE_ACCEPTED_TYPES = {
     "Response",
 }
 
+# JS built-in types that should NOT be wrapped in _BindingWrapper.
+# These have their own Python-side semantics (e.g. passed directly to Response())
+# and wrapping them breaks property access like `.constructor.name`.
+_JS_PASSTHROUGH_TYPES = RESPONSE_ACCEPTED_TYPES | {
+    "Headers",
+}
 
-def _is_response_accepted_type(obj) -> bool:
-    """
-    Check if the given object is an accepted type for a Response body.
-    """
+
+def _get_js_constructor_name(obj) -> str | None:
     if hasattr(obj, "constructor"):
-        return obj.constructor.name in RESPONSE_ACCEPTED_TYPES
-    return False
+        return obj.constructor.name
+    return None
 
 
 class Response(FetchResponse):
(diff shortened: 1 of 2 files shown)


packages/runtime-sdk/src/workers/_workers.py
```diff
@@ -1110,20 +1121,29 @@ class _BindingWrapper:
     def __init__(self, binding):
         self._binding = binding
 
+    def _should_wrap_nested_attribute(self, jsobj) -> bool:
+        if not isinstance(jsobj, JsProxy):
+            return False
+
+        # TODO: This allowlist approach is a workaround. The long-term fix is to
+        # add dedicated Python wrappers for these types in python_from_rpc so they
```
dom96, line 1130: Hmm, I'm not sure that we want Python wrappers for all of these. This fix should be fine though.

dom96 (approved):
Thanks! I actually ran into this yesterday when I upgraded one of my Workers to the latest SDK. Nice to see you caught as well.

packages/runtime-sdk/src/workers/_workers.py
ryanking13: Yeah, we can discuss more about adding wrappers for some types that are used in bindings as an input. I just added that comment so we don't forget, not meaning that we should do it.

PR #129, opened 2026-06-18
feat: add experimental python 314 support

ryanking13:
Adds python 3.14 support on pywrangler. This compat flag is not enabled by date yet, so we are setting the explicit compat flag on our test.


packages/cli/src/pywrangler/utils.py
```diff
@@ -407,6 +413,8 @@ def get_pyodide_index() -> str:
             v = "0.27.7"
         case "3.13":
             v = "0.28.3"
+        case "3.14":
+            v = "314.0.0"
```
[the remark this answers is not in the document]
ryanking13: experimental flag will fail on publish for non-Cloudflare employees so it is safe.

packages/cli/src/pywrangler/utils.py
```diff
@@ -229,7 +229,7 @@ def get_project_root() -> Path:
 
 
 MIN_UV_VERSION = (0, 8, 10)
-MIN_WRANGLER_VERSION = (4, 42, 1)
+MIN_WRANGLER_VERSION = (4, 108, 0)
```
ryanking13: This wrangler version includes python 314 support.

packages/cli/tests/workerd-test/entropy-patches/worker.py
```diff
@@ -21,6 +21,10 @@ async def noop(*args):
 
 class Default(WorkerEntrypoint):
     async def test(self):
+        if sys.version_info >= (3, 14):
+            # FIXME(soon): fix entropy patches for newer packages
```
ryanking13, line 25: pydantic-core issue. We need to fix this, but it should be done in a separate PR.

packages/cli/tests/workerd-test/sdk/tests/test_sdk.py
```diff
@@ -136,17 +136,8 @@ async def test_can_use_undefined_options_and_redirect():
...
         # This tests two things:
         #   * `Response.redirect` static method
-        #   * that other options can be passed into `fetch` (so that we can support
-        #       new options without updating this code)
 
-        # Mock pyodide.http._jsfetch to ensure `foobarbaz` gets passed in.
-        def fetch_check(url, opts):
-            assert opts.foobarbaz == 42
-
-        async with mock_fetch(fetch_check):
```
ryanking13, line 146: I removed this one in our internal workerd tests as well. It mocks pyfetch internal which is easy to break + does not add much value for testing.

packages/cli/src/pywrangler/utils.py
```diff
@@ -239,7 +239,7 @@ def get_pywrangler_config() -> dict:
 
 
 MIN_UV_VERSION = (0, 8, 10)
-MIN_WRANGLER_VERSION = (4, 42, 1)
+MIN_WRANGLER_VERSION = (4, 109, 0)
```
dom96, line 242: I guess you're bumping this to fix the TLS cert issues with Warp, right? If there are other reasons it would be good to document them.

dom96 (approved):
Looks good!

packages/cli/src/pywrangler/utils.py
ryanking13: No it is because each wrangler version comes with a corresponding workerd version and 4.109 contains my python 3.14 change. Let me update the changelog about it.

PR #131, opened 2026-06-18
fix: Ensure that ctx and env __init__ arguments are always wrapped

hoodmane:
We fixed the double-wrapping problem in #126 but added a new problem that ctx and env args in subclasses won't be wrapped at all. This switches to wrapping all __init__ functions but making DurableObjectContext and _EnvWrapper idempotent so that double wrapping them is not possible.

### DIFF
--- a/packages/runtime-sdk/src/workers/_workers.py
+++ b/packages/runtime-sdk/src/workers/_workers.py
@@ -32,8 +32,31 @@ if TYPE_CHECKING:
     from js import DurableObjectState, Env, ExecutionContext
 
 
+def _idempotent_new(cls, obj):
+    """Set __new__ on a class to this so cls is idempotent:
+
+    >>> a = A(x)
+    >>> b = A(a)
+    >>> assert a is b
+
+    For this to work, start the __init__ function with:
+
+    if obj is self:
+        return
+
+    to prevent double-init.
+    """
+    if isinstance(obj, cls):
+        return obj
+    return object.__new__(cls)
+
+
 class _BindingWrapper:
+    __new__ = _idempotent_new
+
     def __init__(self, binding):
+        if binding is self:
+            return
         self._binding = binding
 
     @property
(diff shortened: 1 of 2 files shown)


packages/runtime-sdk/src/workers/_workers.py
```diff
@@ -1610,10 +1613,10 @@ def __init__(self, ctx: "DurableObjectState", env: "Env"):
         self.env = env
 
     def __init_subclass__(cls, **_kwargs):
-        if _is_direct_binding_subclass(cls, DurableObject):
-            _wrap_subclass(cls)
+        _wrap_class(cls)
```
ryanking13, line 553: Do we need both decorator + `__init_subclass__`?

packages/runtime-sdk/src/workers/_workers.py
```diff
@@ -1568,18 +1574,14 @@ def wrapped_init(self, *args, **kwargs):
         original_init(self, *args, **kwargs)
 
     cls.__init__ = wrapped_init
+    return cls
 
 
 def _wrap_workflow_step(cls):
```
ryanking13, line 517: I guess calling `_wrap_workflow_step` twice would still wrap `run` function twice

packages/runtime-sdk/src/workers/_workers.py
```diff
@@ -1627,11 +1630,11 @@ def __init__(self, ctx: "ExecutionContext", env: "Env"):
         self.env = env
 
     def __init_subclass__(cls, **_kwargs: Any):
-        if _is_direct_binding_subclass(cls, WorkerEntrypoint):
-            _wrap_subclass(cls)
-            _wrap_queue_handler(cls)
+        _wrap_class(cls)
+        _wrap_queue_handler(cls)
```
ryanking13, line 571: Same for queue. the `queue` function would be wrapped twice I guess?

packages/runtime-sdk/src/workers/_workers.py
hoodmane: yes

packages/runtime-sdk/src/workers/_workers.py
hoodmane: Oh yeah these need to be adjusted too that is right.

packages/runtime-sdk/src/workers/_workers.py
ryanking13: Okay, this class wrapping thing is really making my head confused

ryanking13 (approved):
Thanks. could you add a testcase that wasn't covered by the previous patch?

hoodmane:
The updates to the test suite did not pass before the change.

PR #132, opened 2026-06-19
fix: fix entropy patch for newer pydantic versions

ryanking13:
Newer pydantic versions removed this function so using newer pydantic versions raises an error (e.g. in python 3.14 test envrionment)

### DIFF
--- a/packages/runtime-sdk/src/_workers_sdk_entropy_import_context.py
+++ b/packages/runtime-sdk/src/_workers_sdk_entropy_import_context.py
@@ -86,6 +86,10 @@ def numpy_random_mtrand_context(module):
 @register_exec_patch("pydantic_core")
 @contextmanager
 def pydantic_core_context(module):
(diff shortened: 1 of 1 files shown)


ryanking13:
Acutally, looks like this is more complicated than this.

packages/runtime-sdk/src/_workers_sdk_entropy_import_context.py
```diff
@@ -86,6 +86,10 @@ def numpy_random_mtrand_context(module):
 @register_exec_patch("pydantic_core")
 @contextmanager
 def pydantic_core_context(module):
+    if not hasattr(module, "validate_core_schema"):
+        # Newer versions of pydantic_core don't have this function
```
hoodmane, line 90: Right you still need to call something that makes pydantic generate a hashmap so the hash seed gets initialized.

PR #133, opened 2026-06-19
chore: add smoke tests for packages that have entropy patches

ryanking13:
Since we are patching these packages. I think it should be useful for us to have test coverage so we don't suddenly break these packages. I found that in newer pydantic-core version (which is available in Python 3.14), our entropy patch doesn't work (#132). I believe these tests will allow us catching those errors when 1) we update Python versions, or when 2) those packages release a new version with breaking changes in entropy.


packages/cli/tests/workerd-test/entropy-patches/tests/test_aiohttp_websocket.py
```diff
@@ -0,0 +1,5 @@
+import aiohttp.http_websocket
+
+
+def test_aiohttp_http_websocket_import():
+    assert aiohttp.http_websocket is not None
```
hoodmane: Does this do anything? Why would this ever be none?

packages/cli/tests/workerd-test/entropy-patches/tests/test_rust_packages.py
```diff
@@ -0,0 +1,48 @@
...
+    "order",
+    list(itertools.permutations(RUST_PACKAGES)),
+    ids=[
+        "->".join(m.split(".")[0] for m in perm)
+        for perm in itertools.permutations(RUST_PACKAGES)
+    ],
+)
+def test_rust_import_permutations(order):
+    for module_name in order:
+        __import__(module_name)
```
hoodmane: This all runs in the same runtime so it's useless.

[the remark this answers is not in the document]
hoodmane: Good bot.

packages/cli/tests/workerd-test/entropy-patches/tests/test_aiohttp_websocket.py
ryanking13: Not very much, actually just importing things should do the validation. Let me update them

dom96 (approved):

PR #136, opened 2026-06-22
fix: ensure self.env and top-level env uses a same class

### LINKED ISSUE #135: RPC methods via workers.env return JsProxy instead of Python types
When calling async methods on a DurableObject via RPC through the `workers.env` binding, the return value is a `pyodide.ffi.JsProxy` (JavaScript Map) instead of being properly deserialized to a Python dict. This breaks type expectations and consistency across different binding access patterns.

Minimal example:

```python
from workers import DurableObject, Response, WorkerEntrypoint, env


class A(DurableObject):
    async def hello(self):
        return {"Hello": "Hola"}


        a = env.A.getByName("a")
        r = await a.hello()
        result = f"{type(r)}: {repr(r)!r}"
        return Response(result, 200)
```

`curl http://localhost:8787/`
>`<class 'pyodide.ffi.JsProxy'>: '[object Map]'`

Expected behavior:

`curl http://localhost:8787/`
>`<class 'dict'>: "{'Hello': 'Hola'}"`

RPC return values should be automatically deserialized from JavaScript types back to their Python equivalents (Map → dict, Array → list, etc.), regardless of whether accessed via `WorkerEntrypoint.env`, `DurableObject.env`, or `workers.env`.

ryanking13:
fixes #135 

Wrap top-level `env` with `_EnvWrapper` so that using `self.env` and top-level `env` goes through the same wrapper.

### DIFF
--- a/packages/runtime-sdk/src/workers/__init__.py
+++ b/packages/runtime-sdk/src/workers/__init__.py
@@ -17,6 +17,7 @@ from ._workers import (
     Response,
     WorkerEntrypoint,
     WorkflowEntrypoint,
+    _EnvWrapper,
     fetch,
     handler,
     import_from_javascript,
@@ -59,7 +60,7 @@ __all__ = [
 def __getattr__(key):
     if key == "env":
         cloudflare_workers = import_from_javascript("cloudflare:workers")
-        return cloudflare_workers.env
+        return _EnvWrapper(cloudflare_workers.env)
     if key in ("wait_until", "waitUntil"):
         cloudflare_workers = import_from_javascript("cloudflare:workers")
         return cloudflare_workers.waitUntil
--- a/packages/runtime-sdk/src/workers/_workers.py
+++ b/packages/runtime-sdk/src/workers/_workers.py
@@ -153,8 +153,45 @@ class FetchKwargs(TypedDict, total=False):
     fetcher: type[pyfetch] | None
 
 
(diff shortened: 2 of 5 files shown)


packages/runtime-sdk/src/workers/__init__.py
```diff
@@ -59,7 +60,7 @@
 def __getattr__(key):
     if key == "env":
         cloudflare_workers = import_from_javascript("cloudflare:workers")
-        return cloudflare_workers.env
+        return _EnvWrapper(cloudflare_workers.env)
     if key in ("wait_until", "waitUntil"):
         cloudflare_workers = import_from_javascript("cloudflare:workers")
         return cloudflare_workers.waitUntil
```
ryanking13, line 66: The waituntil have the same issue (not wrapped with patchWaitUntil), but let's fix it separately.

packages/runtime-sdk/src/workers/_workers.py
```diff
@@ -170,6 +207,10 @@ def body(self) -> "js.ReadableStream":
     def js_object(self) -> "js.Response":
         return self.js_response
 
+    @property
+    def headers(self):
+        return _js_headers_to_http_message(self.js_object.headers)
```
ryanking13, line 212: This change is included to make sure `env.SELF.fetch("...").headers` are kept case-insensitive.

dom96 (approved):

PR #138, opened 2026-06-22
fix: update Workflows wrapper to work more natively with Python objects

ryanking13:
Another bindings update. This update the existing Workflow wrapper classes to take and return native Python objects, so that users does not have to use `to_js` explicitly.

---

__Context for workflows team__: We are updating Python workers SDK so that users does not have to user pyodide FFIs at all, and use the native Python objects directly. After this update, we can remove all the `to_js` usage in workflows docs (https://developers.cloudflare.com/workflows/python/bindings/)


dom96 (approved):
Looks good to me, but would like @Caio-Nogueira to take a look as well.

Caio-Nogueira:
Yeah LGTM, as long as it doesn't break backwards compatibility with current callers that explicitly use `to_js` and `to_py`

PR #141, opened 2026-06-29
chore: factor out util and rpc functions from _workers.py

ryanking13:
This is a mechanical refactor that factors out some common functions from `_workers.py` as this file is getting quite bigger. I will do some more refactor step-by-step.


dom96 (approved):
Nice! Thank you for improving how this code is organised.

PR #142, opened 2026-06-30
chore: factor out request, response, and other type definitions

ryanking13:
More mechanical refactor as a follow-up of #141.

- Type definitions => moved to types.py
- Request => moved to request.py
- Response => moved to response.py
- fetch => moved to fetch.py
- Blob, File => moved to blob.py
- FormData => moved fo formdata.py


packages/runtime-sdk/uv.lock
```diff
@@ -210,7 +210,7 @@ wheels = [
 
 [[package]]
 name = "workers-runtime-sdk"
-version = "1.4.3"
+version = "1.5.1"
```
dom96, line 213: Should this edit be here?

dom96 (approved):
Love it, thanks!

packages/runtime-sdk/uv.lock
ryanking13: Not really, but running uv commands touches this file. I guess we should update our release bot to update this file as well.

PR #144, opened 2026-07-01
Add --allow-build flag & tool.pywrangler setting

hoodmane:
By default we pass `--no-build` because building a binary Pyodide wheel will fail in a confusing way. This makes it annoying to work with local packages though. This adds a `--allow-build` setting to override.

### DIFF
--- a/packages/cli/src/pywrangler/resolve.py
+++ b/packages/cli/src/pywrangler/resolve.py
@@ -18,20 +18,39 @@ MANAGED_SDK_PACKAGE = "workers-runtime-sdk"
 
 
 class InstallPlan:
+    # Lockfile keys that indicate a package is sourced locally
+    _LOCAL_SOURCE_KEYS = ("directory", "sdist", "archive")
+
     def __init__(self, lockfile: Path) -> None:
         self.lockfile = lockfile
         self.requirements: list[tuple[str, str]] = []
+        # Names of packages sourced from a local path. They need refreshing when
+        # rebuilt.
+        self.local_packages: list[str] = []
 
         with open(lockfile, "rb") as f:
             data = tomllib.load(f)
 
         for pkg in data.get("packages", []):
             name = pkg.get("name")
+            if name and any(
+                self._is_local_source(pkg, key) for key in self._LOCAL_SOURCE_KEYS
+            ):
+                self.local_packages.append(name)
+
             version = pkg.get("version")
             if not name or not version:
                 logger.warning("Skipping malformed lockfile entry: %s", pkg)
                 continue
             self.requirements.append((name, version))
+
+    @staticmethod
+    def _is_local_source(pkg: dict, key: str) -> bool:
+        source = pkg.get(key)
+        if not isinstance(source, dict):
+            return False
+        # A local reference has a `path`
+        return "path" in source
 
     def to_requirement_strings(self) -> list[str]:
         return [f"{name}=={version}" for name, version in self.requirements]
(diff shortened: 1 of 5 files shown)


packages/cli/src/pywrangler/resolve.py
```diff
@@ -18,21 +18,40 @@
...
                 logger.warning("Skipping malformed lockfile entry: %s", pkg)
                 continue
             self.requirements.append((name, version))
 
+    @staticmethod
+    def _is_local_source(pkg: dict, key: str) -> bool:
+        source = pkg.get(key)
+        if not isinstance(source, dict):
+            return False
+        # A local reference has a `path``
```
ryanking13:
```suggestion
        # A local reference has a `path`
```

packages/cli/src/pywrangler/resolve.py
```diff
@@ -68,31 +93,38 @@ def _compile_lockfile(
...
-    written to ``pylock.toml``; on subsequent runs the existing file
-    constrains versions so they don't drift.
+    Runs ``uv pip compile`` with the Pyodide interpreter and ``--no-build`` to
+    resolve versions that have Pyodide wheels.  The compiled output is written
+    to ``pylock.toml``; on subsequent runs the existing file constrains versions
+    so they don't drift.
+
+    When *allow_build* is True, ``--no-build`` is omitted so source
+    distributions and local directory sources may be built. Building a binary
+    extension in this way will fail in a confusing way.
```
ryanking13: This comment already exists inside `_compile_lockfile` so I don't think it has to be duplicated here.

packages/cli/tests/test_allow_build.py
```diff
@@ -0,0 +1,269 @@
...
+        requires-python = ">=3.12"
+        dependencies = ["click"]
+        """)
+        + tool_section
+    )
+
+
+# ---------------------------------------------------------------------------
+# Flag / config propagation into sync()
+# ---------------------------------------------------------------------------
```
ryanking13: I think these tests don't add much value. Also, these tests are quite easy to break whenever we update internal logics or parameters.

packages/cli/tests/test_allow_build.py
```diff
@@ -0,0 +1,269 @@
...
+    assert resolve_mock.call_args.kwargs["allow_build"] is True
+
+
+# ---------------------------------------------------------------------------
+# Command construction: --no-build toggling
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize("allow_build", [True, False])
+def test_compile_lockfile_no_build_toggle(tmp_path, allow_build):
```
ryanking13: This too. We are adding too many mocks just to test

```py
if not allow_build:
  args.append("--no-build")
```

which is pretty obvious.

Of course, this kind of test would be great to prevent any kind of regression we make by accident later, but I think adding 4 mock functions just to test this is a code smell. We should our refactor our logic to make the test easier.

packages/cli/tests/test_allow_build.py
ryanking13, line 1: This is missing a real, e2e test.

packages/cli/tests/test_allow_build.py
hoodmane: Yeah agreed this test sucks. I'll hand write something better.

packages/cli/tests/test_cli.py
```diff
@@ -367,6 +371,154 @@ def test_sync_lockfile_lifecycle(test_dir):
...
+    # Without --allow-build: the default --no-build rejects the local source.
+    result = subprocess.run(
+        [*sync_cmd, "--no-allow-build"],
+        capture_output=True,
+        text=True,
+        cwd=test_dir,
+        check=False,
+    )
+    print(f"\n[no-allow-build] stdout:\n{result.stdout}")
+    print(f"[no-allow-build] stderr:\n{result.stderr}")
```
ryanking13: could you please clean up these debug prints before merging?

ryanking13 (approved):
Thanks. I guess we need a better documentation of these CLI flags, but that can be done in a separate PR.

dom96 (approved):

packages/cli/tests/test_cli.py
hoodmane: Aren't debug prints helpful in tests? If the test passes they are hidden, and if the test fails they could be helpful.

hoodmane: Anyways I removed them.

ryanking13: I often use `-rP` flags to debug subprocess errors in tests, and sometimes those debug prints are mixed up and make things harder to debug.

PR #145, opened 2026-07-01
Add wsgi.py analogous to asgi.py

hoodmane:


packages/runtime-sdk/src/workers/wsgi.py
ryanking13, line 1: Considering that we could add more features to these gateway servers, I think it is better to put it in the subdirectory following the regular packages structure (same for asgi.py). We can still have an asgi.py in the top-level directory which just imports from this directory for backward compatibility.

packages/runtime-sdk/src/workers/wsgi.py
```diff
@@ -0,0 +1,278 @@
...
+    def __init__(self, js_body: "js.ReadableStream") -> None:
+        self._reader = js_body.getReader()
+        self._buf = bytearray()
+        self._eof = False
+
+    def readable(self) -> bool:
+        return True
+
+    def _fill(self) -> None:
+        from pyodide.ffi import run_sync
```
ryanking13, line 50: Should we guard against Pyodide 0.26.0a2 which doesn't support run_sync?

packages/runtime-sdk/src/wsgi.py
```diff
@@ -0,0 +1,278 @@
...
+            return 0
+        n = min(len(b), len(self._buf))
+        b[:n] = self._buf[:n]
+        del self._buf[:n]
+        return n
+
+    def close(self) -> None:
+        if not self.closed:
+            try:
+                self._reader.releaseLock()
```
ryanking13: I noticed that workerd raises error when there is a pending read. It is in the [ReadableStream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultReader/releaseLock) spec, but it seems the behavior is different in Node.js / workerd (so I believe AI agents are not very good at it).

so probably this function should make sure there is no pending read by calling `run_sync(await self._reader.cancel()` first.

Also, we need to make sure the lock is released when close is not called.

packages/runtime-sdk/src/wsgi.py
```diff
@@ -0,0 +1,278 @@
...
+            except Exception:  # noqa: BLE001 - best-effort cleanup
+                pass
+        super().close()
+
+
+def _can_stream() -> bool:
+    """Return True if we can lazily stream the body via run_sync"""
+    try:
+        from pyodide.ffi import can_run_sync
```
ryanking13: This should be always true in workerd with Pyodide  >= 0.28.2, otherwise it will not exist at all.

packages/runtime-sdk/src/wsgi.py
```diff
@@ -0,0 +1,278 @@
...
+    """Read the entire request body into memory (fallback when streaming is off)."""
+    if not req.body:
+        return b""
+    chunks = [data.to_bytes() async for data in req.body]
+    return b"".join(chunks)
+
+
+def build_environ(
+    req: "Request | js.Request",
+    env: Any,
```
ryanking13: Maybe give `dict[str, Any]` type?

packages/runtime-sdk/src/wsgi.py
```diff
@@ -0,0 +1,278 @@
...
+    if isinstance(body, (bytes, bytearray)):
+        wsgi_input: Any = io.BytesIO(body)
+        content_length_fallback: int | None = len(body)
+    else:
+        wsgi_input = body
+        content_length_fallback = None
+
+    req_headers = req.headers.items() if isinstance(req, Request) else req.headers
+
+    url = URL.new(req.url)
```
ryanking13: Use urllib instead of js.URL

packages/runtime-sdk/src/wsgi.py
```diff
@@ -0,0 +1,278 @@
...
+    assert url.protocol[-1] == ":"
```
ryanking13: Let's not use assertion in non-test codes. We can properly raise if input is invalid.

packages/runtime-sdk/src/wsgi.py
```diff
@@ -0,0 +1,278 @@
...
+    # `workers.Request.method` is an `http.HTTPMethod` (a str subclass); coerce
+    # to a plain string so frameworks comparing against literals behave.
+    method = str(method.value) if hasattr(method, "value") else str(method)
+
+    environ: dict[str, Any] = {
+        "REQUEST_METHOD": method,
+        "SCRIPT_NAME": "",
+        "PATH_INFO": path_info,
+        "QUERY_STRING": query_string,
+        "SERVER_NAME": url.hostname or "localhost",
```
ryanking13: Will this ever be empty?

packages/runtime-sdk/src/workers/wsgi.py
```diff
@@ -0,0 +1,278 @@
...
+        "wsgi.version": (1, 0),
+        "wsgi.url_scheme": scheme,
+        "wsgi.input": wsgi_input,
+        "wsgi.errors": sys.stderr,
+        "wsgi.multithread": False,
+        "wsgi.multiprocess": False,
+        "wsgi.run_once": False,
+        # Cloudflare-specific extension so handlers can reach bindings, mirroring
+        # the `scope["env"]` convention used by asgi.py.
+        "workers.env": env,
```
ryanking13, line 165: (No action required for this PR) We should properly document this later and never change.

packages/runtime-sdk/src/wsgi.py
```diff
@@ -0,0 +1,278 @@
...
+                if response_state["status"] is not None:
+                    # Headers were already sent; re-raise the original error.
+                    raise exc_info[1].with_traceback(exc_info[2])
+            finally:
+                exc_info = None
+        elif response_state["status"] is not None:
+            raise AssertionError("start_response() called more than once")
+
+        response_state["status"] = status
+        response_state["headers"] = list(response_headers)
```
ryanking13: why do we need to convert to `list`?

packages/runtime-sdk/src/wsgi.py
```diff
@@ -0,0 +1,278 @@
...
+        return write
+
+    result = app(environ, start_response)
+    try:
+        body_chunks = list(write_buffer)
```
ryanking13: This is already a list.

packages/runtime-sdk/src/wsgi.py
```diff
@@ -0,0 +1,278 @@
...
+        # PEP 3333: if the iterable has a close() method, the server must call it.
+        close = getattr(result, "close", None)
+        if close is not None:
+            close()
+
+    if response_state["status"] is None:
+        raise RuntimeError("The WSGI application did not call start_response()")
+
+    return _make_js_response(
+        response_state["status"], response_state["headers"], b"".join(body_chunks)
```
ryanking13: The comment says that the `write_buffer` is only used in the legacy frameworks but it seems we are always using it here.

ryanking13: Also this would store all the data in the memory. I think we should find a way to properly stream it

ryanking13:
Thanks for working on this. I think we need a proper testing strategy with enough coverage before merging this, as we release every commit. Otherwise, we can create a temporary branch that we can incrementally improve until we are confident enough to be merged to main.

packages/runtime-sdk/src/workers/wsgi.py
hoodmane: We do guard.

hoodmane: And I assume you mean 0.26?

ryanking13: > And I assume you mean 0.26?

oh yeah right.

packages/runtime-sdk/src/workers/wsgi.py
```diff
@@ -0,0 +1,383 @@
...
+        ``cancel()`` the reader to drop the remainder; ``cancel()`` returns a JS
+        promise that the runtime settles on its own, so we don't await it.
+        """
+        try:
+            if self._eof:
+                self._reader.releaseLock()
+            else:
+                self._reader.cancel()
+        except Exception:  # noqa: BLE001 - best-effort cleanup
+            pass
```
dom96, line 91: Should we log something here?

dom96 (approved):

ryanking13:
I guess we should be able to this with my Django test suites. If currently uses ASGIHandler, but ideally just replacing it with WSGIHandler would mostly work.

PR #147, opened 2026-07-06
chore: add comprehensive test suites for async django

ryanking13:
Adds comprehensive test suites for async django. It tests common django features and plugins and also DRF, which is a de-facto framework for django. Because of this, the Diff of this PR is pretty big, but most of them are just straightforward (AI-generated) test cases. I'll left a comment for the files that needs some human eye.

Note that django's WSGI-compatible handler is not covered in this PR and #145 needs to be merged to test it.


packages/runtime-sdk/src/asgi.py
```diff
@@ -76,42 +76,86 @@ def request_to_scope(req, env, ws=False):
 
 
 async def start_application(app):
-    shutdown_future = Future()
+    # Drives one ASGI lifespan startup/shutdown cycle before/after serving the
+    # request. The lifespan protocol is optional, so we must tolerate apps that
+    # don't implement it and fall back to serving requests without lifespan.
+    # https://asgi.readthedocs.io/en/latest/specs/lifespan.html
```
ryanking13, line 82: The main change of this PR is I have updated `lifespan` implementation following the spec. django itself does not support lifespan, but we were not gracefully handling it.

packages/runtime-sdk/src/asgi.py
```diff
@@ -256,7 +308,7 @@ def onmessage(evt):
         queue.put_nowait(msg)
 
     server.onopen = onopen
-    server.onopen = onclose
+    server.onclose = onclose
```
ryanking13: Not related to django, but there was a funny typo here.

packages/cli/tests/test_bindings.py
ryanking13, line 1: common functions in this file are moved to conftest.py so we can reuse.

packages/cli/tests/conftest.py
```diff
@@ -3,10 +3,193 @@
...
+
+
+_suite_cache: dict[tuple[str, str], "SuiteResults | str"] = {}
+
+
+def get_test_results(dev_server: str, suite: str) -> SuiteResults:
+    key = (dev_server, suite)
+    if key not in _suite_cache:
+        try:
+            resp = requests.get(f"{dev_server}/run-tests/{suite}", timeout=30)
```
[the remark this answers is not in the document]
ryanking13: Yeah 30 seconds is enough, actually 60 seconds was way too large.

packages/runtime-sdk/src/asgi.py
hoodmane: Can we split it into a PR with the changes to asgi.py + a unit test that covers just the new behavior and a tests-only PR?

ryanking13: Makes sense, will do.

dom96 (approved):
I didn't review this in depth, but had a look at a few and they look good. 

I do wonder though: is there an existing Django test suite that we could/should be running instead?

ryanking13:
I looked at the django codebase to see how they are running unittest, but it was a bit tricky to port them as is as there were additional test requirements, test runners, etc. So I ended up generating tests myself (I gave the codebase to AI and ask it to scan and port the reasonable amount of tests).

ryanking13:
Thanks for the review!

PR #148, opened 2026-07-07
fix: improve lifespan handling in asgi.py

ryanking13:
split out from #147. It only includes asgi.py lifespan + null body handling changes with corresponding tests without django-specific tests.


packages/cli/tests/workerd-test/asgi/worker.py
hoodmane, line 1: Can we make this test use pytest?

ryanking13: Sure.

hoodmane (approved):
Thanks!

packages/cli/tests/workerd-test/asgi/tests/test_asgi.py
```diff
@@ -0,0 +1,384 @@
...
+        if result.done:
+            break
+        body_bytes += result.value.to_bytes()
+
+    expected_size = STREAMING_CHUNK_SIZE * STREAMING_NUM_CHUNKS  # 5120 bytes
+    # This is the key assertion: with bug #10, only the first chunk (1024 bytes)
+    # is returned. The fix should deliver all 5120 bytes.
+    assert len(body_bytes) == expected_size, (
+        f"Expected {expected_size} bytes, got {len(body_bytes)}."
+    )
```
hoodmane: Pytest will do the explanation for us:
```suggestion
    assert len(body_bytes) == expected_size
```

PR #149, opened 2026-07-07
fix: relax required-python version to support Python 3.11

ryanking13:
As discussed internally, this allows installing workers-py in Python 3.11.


packages/runtime-sdk/src/workers/types.py
```diff
@@ -13,11 +13,11 @@ class Context(Protocol):
     def waitUntil(self, other: Awaitable[Any]) -> None: ...
 
 
-type JSBody = (
```
ryanking13, line 16: This syntax is for python >= 3.12.

packages/runtime-sdk/pyproject.toml
```diff
@@ -7,7 +7,7 @@ name = "workers-runtime-sdk"
 version = "1.5.3"
 description = "Python SDK for Cloudflare Workers"
 readme = "README.md"
-requires-python = ">=3.12"
```
ryanking13, line 10: Technically we don't need to update the version requirement of runtime-sdk, but workers-py sets runtime-sdk as a [dev dependency](https://github.com/cloudflare/workers-py/blob/77d279c966df2029711d2d8f6ffebc2610352d8b/packages/cli/pyproject.toml#L22), so this makes our life easier.

hoodmane: Should we remove it from the deps? It doesn't really make sense there?

hoodmane (approved):

packages/runtime-sdk/pyproject.toml
ryanking13: I think it helps IDEs to locate workers-runtime-sdk if it doesn't locate it from the .venv-workers.

PR #150, opened 2026-07-07
chore: rearrange bindings and entrypoints classes

ryanking13:
This is my last code refactor PR:

- Remove `workers/_worker.py`
- Move bindings wrapper implementation to rpc.py
- Move entry points (DurableObjects, Workflows) to a separate entrypoint.py file

### DIFF
--- a/packages/runtime-sdk/src/workers/rpc.py
+++ b/packages/runtime-sdk/src/workers/rpc.py
@@ -6,10 +6,17 @@ from js import Object
 from pyodide.ffi import JsProxy, create_proxy, to_js
 
 from .blob import Blob, File
+from .fetch import fetch
 from .formdata import FormData
 from .request import Request
 from .response import Response
-from .utils import _is_iterable, _is_js_instance, jsnull
+from .utils import (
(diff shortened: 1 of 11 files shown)


packages/runtime-sdk/src/workers/rpc.py
```diff
@@ -206,3 +203,218 @@ def python_to_rpc(value) -> JsProxy:
     )
 
     return result
+
+
+# ---------------------------------------------------------------------------
+# Binding wrappers
```
dom96, line 209: Hmm, this implies to me that this should be in its own module. Perhaps move this into  binding_wrappers.py? or similar

dom96 (approved):
Looks good, but see one suggestion

packages/runtime-sdk/src/workers/rpc.py
ryanking13: Yeah, that was my original implementation, but the problem was that the rpc helper functions and bindings has cross-reference so putting them into a single file can remove the cyclic imports. Let's put them as-is for now and think about further refactor later.

PR #153, opened 2026-07-09
fix: fix entropy patch for newer pydantic versions

ryanking13:
This is on top of #129. It fixes pydantic-core entropy patch so that we can use fastapi in python 3.14


packages/runtime-sdk/src/_workers_sdk_entropy_import_context.py
```diff
@@ -92,14 +92,22 @@ def pydantic_core_context(module):
         with allow_bad_entropy_calls(1):
             yield
     finally:
-        try:
+        validate_core_schema = getattr(module, "validate_core_schema", None)
+        if validate_core_schema is None:
+            # pydantic-core >=2.41 no longer exports validate_core_schema. Build a
+            # tiny validator instead so any lazy hash-map seeds are initialized
```
hoodmane:
```suggestion
            # Build a tiny validator instead so any lazy hash-map seeds are initialized
```

hoodmane: Can we just delete the other branch? This should work always.

ryanking13: yeah, that would work too.

joesepi (approved):

hoodmane (approved):

PR #130, opened 2026-06-18
feat: wrap AI, Images, Analytics Engine, Vectorize and RateLimit Bindings to accept native Python objects

ryanking13:
Now that we have a good BindingWrapper class, these just work. Unfortunately AI and Vectorize bindings require a real prod account even for the local testing, so we cannot run unittest in CI.

I verified them working my making a demo worker with my local account + patched sdk.

<details><summary>Vectorize test worker</summary>
<p>

This requires 64-dimension DB in the account

```python
import traceback

from workers import Response, WorkerEntrypoint


class Default(WorkerEntrypoint):
    async def fetch(self, request):
        from urllib.parse import urlparse

        path = urlparse(request.url).path

        if path == "/probe":
            vec = self.env.VECTORIZE
            return Response.json({
                "type": str(type(vec)),
                "wrapped": hasattr(vec, "_binding"),
                "has_insert": hasattr(vec, "insert"),
                "has_query": hasattr(vec, "query"),
            })

        if path == "/describe":
            try:
                result = await self.env.VECTORIZE.describe()
                return Response.json({"success": True, "result": result})
            except Exception as e:
                return Response.json({
                    "success": False,
                    "error": str(e),
                    "type": type(e).__name__,
                    "traceback": traceback.format_exc(),
                }, status=500)

        if path == "/insert":
            try:
                vectors = [
                    {"id": "vec-1", "values": [0.1] * 64},
                    {"id": "vec-2", "values": [0.4] * 64},
                ]
                result = await self.env.VECTORIZE.insert(vectors)
                return Response.json({"success": True, "result": result})
            except Exception as e:
                return Response.json({
                    "success": False,
                    "error": str(e),
                    "type": type(e).__name__,
                    "traceback": traceback.format_exc(),
                }, status=500)

        if path == "/query":
            try:
                result = await self.env.VECTORIZE.query(
                    [0.1] * 64,
                    {"topK": 5},
                )
                return Response.json({"success": True, "result": result})
            except Exception as e:
                return Response.json({
                    "success": False,
                    "error": str(e),
                    "type": type(e).__name__,
                    "traceback": traceback.format_exc(),
                }, status=500)

        if path == "/upsert":
            try:
                vectors = [
                    {"id": "vec-3", "values": [0.7] * 64, "metadata": {"label": "test"}},
                ]
                result = await self.env.VECTORIZE.upsert(vectors)
                return Response.json({"success": True, "result": result})
            except Exception as e:
                return Response.json({
                    "success": False,
[...]

PR #155, opened 2026-07-16
fix: wrap AnalyticsEngine binding

### LINKED ISSUE #154: [Workers Analytics Engine] Type AnalyticsEngine is not wrapped correctly.
Demo:

```python
from workers import WorkerEntrypoint, Response


        try:
            self.env.ANALYTICS.writeDataPoint(
                {
                    "blobs": ["blob1", "blob2"],
                    "doubles": [1.0, 2.5],
                    "indexes": ["idx"],
                }
            )

        except Exception as inst:
            return Response(str(inst))

        return Response("Hello world!")

```

After `wrangler deploy`, curl says:

```
TypeError: Incorrect type for the 'indexes' field on 'AnalyticsEngineEvent': the provided value is not of type 'Array'.
```

VergeDX:
Close #154

### DIFF
--- a/packages/runtime-sdk/src/workers/_workers.py
+++ b/packages/runtime-sdk/src/workers/_workers.py
@@ -251,6 +251,7 @@ class _EnvWrapper:
         "WorkerQueue",
         "Ai",
         "VectorizeIndexImpl",
+        "AnalyticsEngine",
         "AnalyticsEngineDataset",
         "LocalAnalyticsEngineDataset",
         "ImagesBindingImpl",


VergeDX:

dom96 (commented):
Thank you for this! Any chance you could add a quick test for this too?

VergeDX:
@dom96 Actually, we already have tests here:
https://github.com/cloudflare/workers-py/blob/main/packages/cli/tests/bindings-test/src/test_analytics_engine.py
But the original issue can only be reproduced in the production environment (Requires `wrangler deploy`).

This is because `wrangler dev` (local server) uses a mock implementation from Miniflare:
https://github.com/cloudflare/workers-sdk/blob/main/packages/miniflare/src/workers/analytics-engine/analytics-engine.worker.ts

And seems the CI tests also run on the local server:
https://github.com/cloudflare/workers-py/blob/main/packages/cli/tests/test_bindings.py

VergeDX:
Ping @ryanking13 because #130 adds support for the Workers Analytics Engine.

ryanking13:
Yeah, it is tricky to test this as these classes sometimes have different class names in production and workerd dev environment. Thanks for catching this

ryanking13 (approved):
Looks good. Could you update the commit message to make the CI happy? 

```
⧗   input: fix: Wrap AnalyticsEngine, which is used in production.
✖   subject may not end with full stop [subject-full-stop]
```

PR #159, opened 2026-07-29
docs: run the pinned ruff via pre-commit in CONTRIBUTING

whitphx:
Two problems with the current instruction:

1. A bare `uvx ruff` runs the latest ruff release, while CI enforces the version pinned in `.pre-commit-config.yaml` (currently `v0.9.1`). This can lead to inconsistent formatting and linting between the CI and local environments.
2. `uvx ruff fix .` looks like a typo which is not a ruff subcommand (the linter's fix mode is `ruff check --fix`).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,7 +54,18 @@ Each package includes these development tools:
    cd packages/<package-name>
    ```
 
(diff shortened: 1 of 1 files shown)


ryanking13 (approved):
Thanks!

PR #160, opened 2026-07-29
fix: fix function object returned from RPC to be a callable

ryanking13:
This fixes some bugs related to RPC calls, that I noticed while checking [RPC docs examples](https://developers.cloudflare.com/workers/runtime-apis/rpc/).

1. When RPC call returns a function (closure), it was not interpreted as callable.
2. calling `bool(binding)` was failing because the Python class fallbacks to `len()` call when there is no `__bool__` implemented.

### DIFF
--- a/packages/runtime-sdk/src/workers/rpc.py
+++ b/packages/runtime-sdk/src/workers/rpc.py
@@ -286,7 +286,7 @@ class _BindingWrapper:
         js_kwargs = {k: python_to_rpc(v) for k, v in kwargs.items()}
         return js_args, js_kwargs
 
-    def _invoke(self, fn, args, kwargs):
+    def _conver_and_invoke(self, fn, args, kwargs):
         js_args, js_kwargs = self._convert_args(args, kwargs)
         result = fn(*js_args, **js_kwargs)
         if hasattr(result, "then") and callable(result.then):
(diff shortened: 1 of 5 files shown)


packages/runtime-sdk/src/workers/rpc.py
```diff
@@ -255,45 +255,56 @@ def _should_wrap_nested_attribute(self, jsobj) -> bool:
...
+            return [self._wrap_js_value(item) for item in converted]
+        return self._wrap_js_value(converted)
 
     @staticmethod
     def _convert_args(args, kwargs):
         js_args = [python_to_rpc(arg) for arg in args]
         js_kwargs = {k: python_to_rpc(v) for k, v in kwargs.items()}
         return js_args, js_kwargs
 
+    def _invoke(self, fn, args, kwargs):
```
dom96: nit: maybe name this `_convert_and_invoke`

dom96 (approved):
Weird, I thought I tested this before and it worked. Maybe a regression?

Anyway, looks good, thanks!

joesepi (approved):

packages/runtime-sdk/src/workers/rpc.py
ryanking13: sounds good

ryanking13:
> Weird, I thought I tested this before and it worked. Maybe a regression?

Yeah, I think it was a regression after I added BindingsWrapper class. Previously it should be a raw js proxy.

PR #163, opened 2026-07-30
chore: test django with WSGI

ryanking13:
Run the same django test suite that I added in #147 with wsgi.py


packages/cli/tests/web-frameworks-test/django-async/src/test_caching.py
ryanking13, line 1: renamed to `/django/src/test_caching.py`. It seems like github is not smart enough to detect renaming on this file.

packages/cli/tests/web-frameworks-test/django-async/src/test_streaming.py
ryanking13, line 1: Same. renamed to `/django/src/test_streaming.py`.

packages/runtime-sdk/tests/web-frameworks-test/django/src/django_app/middleware.py
```diff
@@ -1,16 +1,24 @@
 from asgiref.sync import iscoroutinefunction, markcoroutinefunction
 
 
-class CustomAsyncMiddleware:
+class CustomMiddleware:
```
ryanking13, line 4: Making this middleware work on both sync and async mode.

packages/runtime-sdk/tests/web-frameworks-test/django/src/django_app/urls_sync.py
ryanking13, line 1: django's async and sync views need to be registered differently. The only difference between `urls.py` and `urls_sync.py` is that whether the views are synchronous or asynchronous (async def  vs  def)

packages/runtime-sdk/src/workers/wsgi.py
ryanking13, line 1: This change is splitted off to #162

PR #164, opened 2026-07-30
fix(runtime-sdk): deliver binary WebSocket frames as ASGI `bytes`

whitphx:
This PR takes care of the case where the WebSocket payload is binary, not text: `onmessage` wrapped every frame under the ASGI `text` key, so frameworks' `receive_bytes()` paths never fired. Binary frames are now enqueued under `bytes`, and the server socket sets `binaryType = "arraybuffer"` so the frame can be read synchronously in the message callback.

`pytest -k asgi-ws` in `packages/runtime-sdk` runs the workerd suite, which asserts that a text frame and a binary frame each arrive under the right ASGI key.

### DIFF
--- a/packages/runtime-sdk/src/asgi.py
+++ b/packages/runtime-sdk/src/asgi.py
@@ -294,6 +294,9 @@ async def process_websocket(app: Any, req: "Request | js.Request") -> js.Respons
 
     client, server = WebSocketPair.new().object_values()
     server.accept()
+    # Binary frames otherwise arrive as Blob (observed under wrangler dev),
+    # which cannot be read synchronously in the message callback.
+    server.binaryType = "arraybuffer"
     queue = Queue()
 
     def onopen(evt):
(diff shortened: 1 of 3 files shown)


ryanking13 (commented):
Thanks! codewise looks good to me but could you please update existing workerd-test tests instead of creating a new directory?

We have quite a lot of boilerplate code to run workerd test + each test requires invoking workerd which is expensive, so we try to reduce the number of test directories (same for other PRs you created).

whitphx:
Thank you for the comment. PR updated.

ryanking13:
It seems like timeouts are real. Could you please check if you can reproduce the timeout locally? I guess connecting to websocket hangs somehow

ryanking13:
There was an outage in GitHub actions today, which made the CI didn't run I guess.

whitphx:
I closed and re-opened the PR to trigger the CI. Can you see the approval button now?

ryanking13 (approved):
Thanks!

PR #165, opened 2026-07-30
fix(runtime-sdk): send empty WebSocket frames instead of dropping them

whitphx:
Care the case where the WebSocket payload is an empty text or empty binary frame: `ws_send` tested the payload for truthiness, so `b""` and `""` were skipped and never reached the client. Both keys are now compared against `None`, which is the only value that means "not this kind of frame".

`pytest -k asgi-ws` in `packages/runtime-sdk` runs the workerd suite. The test covers the app-to-client direction, where the frames are dropped, and times out without the fix.


hoodmane:
So you're saying this already had test coverage and the tests were broken?

whitphx:
No, I meant the `if`-clauses in `ws_send()` were not perfect and I just fixed them and added test cases for it in this PR.

ryanking13 (approved):
Thanks!

PR #166, opened 2026-07-30
fix(runtime-sdk): close the WebSocket transport when the app task ends

whitphx:
When the ASGI app task ends after accept without sending `websocket.close`, the transport is left open, so the client keeps a half-open connection and hangs instead of reconnecting. Closing the connection once the app is done is the server's job, so the server now closes it: 1011 when the task raised, 1000 otherwise.

The task is also registered with the runtime through `wait_until`, as `process_request` already does, so its lifetime after the 101 rests on the platform contract rather than on local dev happening to keep orphan tasks alive.

`pytest -k asgi-ws` in `packages/runtime-sdk` runs the workerd suite, which asserts the client sees a 1011 close after the app crashes.

### DIFF
--- a/packages/runtime-sdk/src/workers/asgi.py
+++ b/packages/runtime-sdk/src/workers/asgi.py
@@ -16,7 +16,7 @@ logger = logging.getLogger("asgi")
 background_tasks = set()
 
 
-def run_in_background(coro: Awaitable[Any]) -> None:
+def run_in_background(coro: Awaitable[Any]) -> Future:
     fut = ensure_future(coro)
     background_tasks.add(fut)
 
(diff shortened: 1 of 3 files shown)


packages/runtime-sdk/src/asgi.py
```diff
@@ -368,7 +375,36 @@ async def ws_receive():
...
+            # otherwise the client keeps a half-open connection and hangs
+            # instead of reconnecting.
+            try:
+                server.close(1011 if exc is not None else 1000, "")
+            except Exception:
+                pass  # the peer may already have closed the socket
+        task_proxy.destroy()
+
+    task.add_done_callback(_on_done)
+    wait_until(task_proxy)
```
ryanking13: I don't like that we are basically copy-pasting `run_in_background` here with some additional features. Maybe we could refactor `run_in_background` to use wait_until internally to make sure the background works can be done before the isolate is terminated?

ryanking13 (approved):
Thanks! maybe we should improve `wait_until` to wrap functions with create_proxy by default.

PR #167, opened 2026-07-30
fix(runtime-sdk): preserve repeated response headers (multiple `Set-Cookie`)

whitphx:
Building a header object with `Object.fromEntries(headers)` drops repeated header contents.
For example, when a response endpoint returns the header like this,
```
Set-Cookie: session=eyJhbGciOi...; Path=/; HttpOnly; Secure
Set-Cookie: csrftoken=a8f5f167f4...; Path=/; Secure; SameSite=Lax
```

Through the current bridge, `Object.fromEntries` collapses the pair and the client receives only:
```
Set-Cookie: csrftoken=a8f5f167f4...; Path=/; Secure; SameSite=Lax
```

This PR fixes it by building the headers with the SDK's own `_to_js_headers` helper, which keeps them as a `js.Headers`.

`pytest -k asgi` in `packages/runtime-sdk` runs the workerd suite, which asserts both cookies survive the round trip.

### DIFF
--- a/packages/runtime-sdk/src/asgi.py
+++ b/packages/runtime-sdk/src/asgi.py
@@ -8,6 +8,7 @@ from urllib.parse import unquote
 import js
 
 from workers import Context, Request
+from workers.utils import _to_js_headers
 
 ASGI = {"spec_version": "2.0", "version": "3.0"}
 NULL_BODY_STATUSES = frozenset({101, 103, 204, 205, 304})
@@ -179,7 +180,7 @@ async def process_request(
     # TODO(later): remove this parameter after unvendoring Python SDK from workerd
     ctx: Context | None,
 ) -> js.Response:
-    from js import Object, Response, TransformStream
+    from js import Response, TransformStream
     from pyodide.ffi import create_proxy
 
     status = None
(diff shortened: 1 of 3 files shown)


packages/runtime-sdk/src/asgi.py
```diff
@@ -40,6 +40,16 @@ def acquire_js_buffer(pybuffer):
...
 
 
+def _to_js_headers(headers):
+    # js.Headers preserves repeated names per the Fetch spec; building the
+    # response headers with Object.fromEntries collapses them to the last
+    # value per name, which silently drops all but one Set-Cookie.
+    from js import Headers
+    from pyodide.ffi import to_js
+
+    return Headers.new(to_js(headers))
```
ryanking13: Could you please check if you can use [_to_js_headers](https://github.com/cloudflare/workers-py/blob/88f53c0198d6af4438925ddf98e5197a0a5513b4/packages/runtime-sdk/src/workers/utils.py#L172) in utils.py directly? I guess `to_js` is also not needed.

ryanking13 (commented):
Thanks! I left one comment about reusing the existing code, otherwise looks good to me.

packages/runtime-sdk/tests/workerd-test/asgi/tests/test_asgi.py
```diff
@@ -399,3 +399,10 @@ async def test_scope_path_is_percent_decoded():
 async def test_scope_exposes_raw_path():
     scope = await _scope("/scope/hello%20world")
     assert scope["raw_path"] == "/scope/hello%20world"
+
+
+@pytest.mark.asyncio
+async def test_multiple_set_cookie_headers_survive():
+    response = await env.SELF.fetch("http://example.com/multi-cookie")
+    cookies = list(response.js_object.headers.getSetCookie())
```
ryanking13: Let's check `response.headers` property instead, which is a Pythonic wrapper. `js_object` is not meant to be used by end users normally.

ryanking13 (approved):

PR #168, opened 2026-07-30
fix(runtime-sdk): terminate the response stream when the app fails mid-stream

whitphx:
When the app raises after a streaming response has started, the `TransformStream` writer is left open, so the client waits forever for an EOF that never arrives. This PR ends the response stream on that path, so the client sees the truncated body instead of hanging.

The close is dispatched through `run_in_background` rather than awaited, because `writer.close()` settles only once the queued chunks reach a consumer: awaiting it for a response whose body nobody reads would block the app task that the runtime waits on through `wait_until`, and the request would be cancelled as hung. A pending `receive()` is woken with `http.disconnect` at the same point.

`pytest -k asgi` in `packages/runtime-sdk` runs the workerd suite, which reads the stream to completion after the app fails mid-body.

### DIFF
--- a/packages/runtime-sdk/src/asgi.py
+++ b/packages/runtime-sdk/src/asgi.py
@@ -28,6 +28,13 @@ def run_in_background(coro: Awaitable[Any]) -> None:
     fut.add_done_callback(_on_done)
 
 
+async def close_stream_quietly(writer):
+    try:
+        await writer.close()
+    except Exception:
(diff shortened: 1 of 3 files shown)


ryanking13 (approved):

ryanking13:
Hmm, actually this seems to be breaking one of our existing test.

whitphx:
Thank you for the review. I will take a look.

ryanking13 (approved):

PR #169, opened 2026-07-30
fix(runtime-sdk): percent-decode `scope["path"]` per the ASGI spec

whitphx:
`scope["path"]` was passed percent-encoded, but the ASGI spec requires the decoded path there (the encoded form belongs in `raw_path`). For example, a request to `/files/my%20file.txt` reached routing as `/files/my%20file.txt` instead of `/files/my file.txt`.

`pytest -k asgi` in `packages/cli` runs the tests, which echo both scope fields back for a percent-encoded request.


ryanking13 (approved):
Thanks!

PR #170, opened 2026-07-30
fix(runtime-sdk): strip `accept-encoding` from the scope headers

whitphx:
The Workers runtime compresses responses itself, so forwarding the client's `accept-encoding` to the app lets app-level compression middleware (e.g. Starlette's `GZipMiddleware`) produce double-compressed bodies. This PR strips it from the scope.

`pytest -k asgi` in `packages/runtime-sdk` runs the workerd suite. The test sends a second custom header as a control, so the assertion cannot pass vacuously.


ryanking13:
I am not sure if this is a right fix. The users should not use the middleware in that case if they are aware of that the server will do the compression already.

hoodmane:
It kind of makes sense to me.

whitphx:
Thank you for the review, and you are correct. Doing this is too much for the SDK layer. Withdraw this PR.

PR #172, opened 2026-07-30
fix(runtime-sdk): stop losing details of ASGI lifespan failures

whitphx:
Two lifespan failure paths discard what went wrong. An empty but present `message` produced `RuntimeError('')`, because a `.get` default only applies when the key is absent. An exception raised before the startup ack was swallowed along with the spec-mandated "this app has no lifespan handler" conclusion, so a crash partway through startup work looked identical to an app that never implemented lifespan; it is now logged at debug level.

### Test Plan

```
$ uv run pytest 'tests/test_in_workerd.py::test_in_workerd[asgi-3.13]' -v
```

(run from `packages/runtime-sdk`)

### DIFF
--- a/packages/runtime-sdk/src/workers/asgi.py
+++ b/packages/runtime-sdk/src/workers/asgi.py
@@ -127,7 +127,9 @@ async def start_application(app):
                 startup.set_result(True)
             return
         if got["type"] == "lifespan.startup.failed":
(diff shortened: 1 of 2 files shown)


ryanking13 (approved):
Makes sense. Thanks!

PR #181, opened 2026-08-05
TEMP DO NOT REVIEW: probe rebaseable flag (will be closed)

ryanking13:
Temporary probe to determine why PR #180 reports `rebaseable: false`.

Identical to #180 except the four binary `uv.lock` files are excluded from history.
Will be closed as soon as the API flag is read. Do not review or merge.


ryanking13:
Probe complete. Root cause: GitHub's 100-commit rebase limit (confirmed 100=rebaseable, 101=not). Closing.

PR #183, opened 2026-08-06
chore(django-cf): apply ruff to the codebase

ryanking13:
Apply the same linter settings to other two packages in this repository. This change is mostly mechanical and NFC.

- Apply ruff (mypy and semgrep is not applied yet)
- Test directories are excluded for now.


dom96 (approved):
Can we run `ruff` in CI too so we can ensure we don't need a PR like this again?

ryanking13:
> Can we run `ruff` in CI too so we can ensure we don't need a PR like this again?

Yes, this PR enables it. It updates pre-commit config which disabled ruff previously for this subdirectory.

PR #186, opened 2026-08-06
fix(django-cf): fix binding results type conversions

ryanking13:
Removed extra `to_py()` or other conversion logics from django-cf. I have updated our rpc calls in sdk to do these conversions internally so that users will only see the Python objects whenever possible.

I also dropped the DO-backend behavior that were exposing JS stacks to the user, which is not useful.

These changes are mostly to make the existing unittests pass, not to actually introduce new features.


ryanking13:
> One actionable issue found: Three .to_py() calls remain in r2.py

These return JS classes, which works like a dict, so I think we can keep them as-is for now.

ryanking13:
Will reopen after rebase

PR #187, opened 2026-08-06
fix(django-cf): run the django-cf suite in CI

ryanking13:
Runs existing django-cf tests in CI. We will probably need to adopt some of the test fixtures from runtime-sdk as a follow up.


dom96 (approved):

ryanking13:
I ended up refactoring the test suite a bit to align with our current unittest structure. Let me clean things up as a follow up but merging for now so that other PRs can land too.

PR #190, opened 2026-08-07
Implements tests for FastAPI StaticFiles and anyio thread patch.

dom96:
Implements StaticFile tests which fail without the patch. Will follow up with more tests.

Uses .pth mechanism to patch anyio to fix the "can't start new thread" errors.


hoodmane (approved):
Thanks!

ryanking13 (approved):
Thanks! Let's see if this works well. At least this is far better than showing "cannot start a new thread" :)

PR #191, opened 2026-08-07
fix: Make wsgi streaming response handlers async

hoodmane:
So it's possible to stack switch inside of them. Also, make sure that contextvars are preserved.

On top of #189.

### DIFF
--- a/packages/runtime-sdk/src/workers/wsgi.py
+++ b/packages/runtime-sdk/src/workers/wsgi.py
@@ -262,8 +262,9 @@ def _make_streaming_response(
             for proxy in proxies:
                 proxy.destroy()
 
+    # Make proxies async so that it is possible to stack switch inside them.
     @create_proxy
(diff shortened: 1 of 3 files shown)


packages/runtime-sdk/src/workers/wsgi.py
```diff
@@ -250,22 +251,24 @@ def _make_streaming_response(
...
             for proxy in proxies:
                 proxy.destroy()
 
+    # Make proxies async so that it is possible to stack switch inside them.
     @create_proxy
-    def pull(controller: Any) -> None:
+    async def pull(controller: Any) -> None:
         try:
-            chunk = next(chunks, _END)
+            chunk = ctx.run(next, chunks, _END)
```
ryanking13: So if we don't set this the contextvars can be leaked between requests? Does this only apply to the streaming response?

ryanking13 (approved):
Looks good to me. I left one question.

PR #192, opened 2026-08-10
chore: adds more fastapi threading tests

dom96:
Just a few more test cases which end up creating a thread via starlette:

* Sync routes
* Streaming
* File upload

Follow up to https://github.com/cloudflare/workers-py/pull/190


ryanking13 (approved):
Nice!

PR #194, opened 2026-08-12
chore(runtime-sdk): relocate asgi.py into workers directory for consistency

ryanking13:
moving `asgi.py` to `workers/asgi.py` so that users can import both wsgi and asgi modules from

```py
from workers import asgi
from workers import wsgi
```

I made the existing asgi.py work as-is by importing from workers/asgi.py and re-exporting.


packages/runtime-sdk/src/asgi.py
```diff
@@ -1,407 +1,6 @@
...
-        from fastapi import Depends, Request
-
-        @Depends
-        async def env(request: Request):
-            return request.scope["env"]
-
-        return env
-
-    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+sys.modules[__name__] = _asgi
```
dom96, line 6: Should we print a deprecation warning here?

dom96 (approved):

packages/runtime-sdk/src/asgi.py
ryanking13: Since the examples, documents and existing packages are all importing from `import asgi` directly. I think it would be too noisy to do it right now. Maybe let's do it after we replace all the examples and existing code not to import from top-level asgi directly.

PR #196, opened 2026-08-12
ci(django-cf): release django-cf from ci

ryanking13:
Apply the same release process that other packages in this repository use to django-cf, so that the package can be released by commit message.


dom96 (approved):
Does this auto release all packages in the repo, even if only one had changes? or is it smart enough to know to only release the packages that changed?

ryanking13:
Only the package that the code inside the package directory is modified is released (filtered by `path_filters` key)

PR #197, opened 2026-08-12
Add default enable date for 3.14 and remove experimental

dom96:
We could really use some way to ensure we don't forget this as it does cause issues.

### DIFF
--- a/packages/cli/src/pywrangler/metadata.py
+++ b/packages/cli/src/pywrangler/metadata.py
@@ -10,7 +10,9 @@ class PythonCompatVersion(NamedTuple):
 
 
 PYTHON_COMPAT_VERSIONS = [
-    PythonCompatVersion("3.14", "python_workers_20260610", None, experimental=True),
+    PythonCompatVersion(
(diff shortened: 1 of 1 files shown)


dom96:
Though when I do get get_python_version to return 3.14, I get: `error: No interpreter found for cpython-3.14.2-emscripten-wasm32-musl in managed installations or search path`

Of course with 3.13 I get the classic `node: bad option: --experimental-wasm-stack-switching`

ryanking13:
> Though when I do get get_python_version to return 3.14, I get: error: No interpreter found for cpython-3.14.2-emscripten-wasm32-musl in managed installations or search path

Hmm that's weird. What do you get when you run 

```bash
uv python list | grep wasm
```

?

ryanking13 (approved):
Codewise looks good

dom96:
```
$ uv python list | grep wasm
pyodide-3.13.2-emscripten-wasm32-musl               /Users/dom/.local/share/uv/python/pyodide-3.13.2-emscripten-wasm32-musl/python
```

ryanking13:
Looks like uv is not properly installing python interpreter for some reason... 🤔

PR #198, opened 2026-08-12
test(django-cf): use real d1 and durable objects instead of mocks

ryanking13:
Previously d1 and durable objects tests were using mock Python objects. This updates it to use a real local binding provided by wrangler, which makes the test more robust.


dom96 (approved):
Nice!

PR #202, opened 2026-08-13
fix(django-cf): fix order_by not working properly

ryanking13:
Fixes a bug that the fetch order is always fixed for the d1, do backend.

### DIFF
--- a/packages/django-cf/django_cf/db/base_engine.py
+++ b/packages/django-cf/django_cf/db/base_engine.py
@@ -201,6 +201,7 @@ class CFResult:
 
     def __init__(self, data):
         self.data = data
+        self._position = 0
 
     def __iter__(self):
         return iter(self.data)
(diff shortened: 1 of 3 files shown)


packages/django-cf/django_cf/db/base_engine.py
```diff
@@ -212,29 +212,23 @@ def set_rowcount(self, value):
         self.rowcount = value
 
     def fetchone(self):
-        if len(self.data) > 0:
-            return self.data.pop()
+        if self._position < len(self.data):
```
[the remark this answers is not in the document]
ryanking13: fixed

packages/django-cf/django_cf/db/base_engine.py
```diff
@@ -212,29 +213,23 @@ def set_rowcount(self, value):
...
-                break
-            ret.append(row)
-            if size is not None:
-                size -= 1
-
+        if size <= 0:
+            return []
+        end = min(self._position + size, len(self.data))
+        ret = self.data[self._position : end]
+        self._position = end
```
dom96, line 232: Curious about the changes here, why make them? The old code looks fine?

dom96 (approved):
Curious why the changes to packages/django-cf/django_cf/db/base_engine.py are needed, but otherwise looks good.

packages/django-cf/django_cf/db/base_engine.py
ryanking13: Fixing + performance.

1. Originally it was popping from the back, creating the reverse-ordered result.
2. slicing rather than pop + append

PR #205, opened 2026-08-14
fix(django-cf): return a fresh new cursor for every cursor call

ryanking13:
Fixes a bug that calling `cursor()` against a database connection was returning always the same cursor.

```py
cursorA = connection.cursor()
cursorB = connection.cursor()

assert cursorA != cursorB  # should not be the same!
```

### DIFF
--- a/packages/django-cf/django_cf/db/base_engine.py
+++ b/packages/django-cf/django_cf/db/base_engine.py
@@ -290,26 +290,21 @@ class CFCursor:
         self.database.defer_foreign_keys(state)
 
     def fetchone(self):
-        result = self.lastResult
-        return result.fetchone()
+        return self.lastResult.fetchone()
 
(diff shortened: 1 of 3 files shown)


packages/django-cf/django_cf/db/base_engine.py
```diff
@@ -277,63 +277,41 @@ def from_object(
...
-        return  # No commits allowed
-
-    def rollback(self):
-        return  # No commits allowed
+        self.database.defer_foreign_keys(state)
 
     def fetchone(self):
-        return self.lastResult.fetchone()
+        result = self.lastResult
+        return result.fetchone()
```
dom96: Nit: for a cleaner diff you can avoid `result` here and below

dom96 (approved):

PR #207, opened 2026-08-17
chore: implements response type tests for fastapi

dom96:


packages/runtime-sdk/tests/web-frameworks-test/fastapi-tests/src/test_response_types.py
```diff
@@ -0,0 +1,183 @@
...
+
+@pytest.mark.asyncio
+async def test_set_cookie(fastapi_app):
+    """Response.set_cookie emits a Set-Cookie header."""
+    resp = await fetch(fastapi_app, "/responses/set-cookie")
+    assert resp.status == 200
+    set_cookie = resp.headers.get("set-cookie")
+    assert set_cookie is not None
+    assert "session_id=abc123" in set_cookie
+    assert "httponly" in set_cookie.lower()
```
ryanking13, line 106: Maybe it would be nice to test passing multiple set-cookie headers, as set-cookie needs to be treated specially. I believe we have a test for it in asgi.py itself, but would be nice to test it in real frameworks too.

ryanking13 (approved):
Left one comment, otherwise looks good

packages/runtime-sdk/tests/web-frameworks-test/fastapi-tests/src/test_response_types.py
dom96: Will make a note to add a separate test for this

PR #211, opened 2026-08-18
fix(cli): locate pyodide vendor site-packages by discovery, not a guessed OS-conditional path

### LINKED ISSUE #212: Python Workers: ModuleNotFoundError: No module named 'workers' on deploy, even with official pywrangler init scaffold
### What versions & operating system are you using?

System:
  OS: Windows 10 10.0.19045
  CPU: Intel Core i5-4300U
Binaries:
  Node: 24.18.0
  npm: 11.16.0
npmPackages (worker project):
  wrangler: 4.123.0
  workers-py: 1.16.4
  workers-runtime-sdk: 1.6.13

### Please provide a link to a minimal reproduction

No hosted repo needed - reproduces with the official scaffold itself: `uvx --from workers-py pywrangler init myapp --accept-defaults --no-deploy --no-git --no-open` then `cd myapp && uv run pywrangler deploy` (unmodified generated entry.py/pyproject.toml/wrangler.jsonc).

### Describe the Bug

Deploying any Python Worker fails at the Cloudflare API validation step with a ModuleNotFoundError for the 'workers' package - even the entry.py generated by the official scaffold.

Steps to reproduce:
1. uvx --from workers-py pywrangler init myapp --accept-defaults --no-deploy --no-git --no-open
2. 2. cd myapp
3. 3. uv run pywrangler deploy
Expected: deploy succeeds with the unmodified generated template.
Actual: fails with the error pasted below. The traceback shows it originates from Cloudflare's own internal introspection.py trying `from workers import (...)`, not from user code - the generated entry.py's own `from workers import Response, WorkerEntrypoint` line is where it dies.

I also tried a hand-written minimal Worker (single file, `async def on_fetch(request, env)` + `js.Response.new(...)`, zero dependencies, wrangler.toml with only compatibility_flags = ["python_workers"]) via plain `wrangler deploy` - identical error.

Things I tried that did not fix it:
- compatibility_flags: ["python_workers", "disable_python_external_sdk"] - gets past this specific error, but then fails differently with "The uploaded script has no registered event handlers" [code: 10068].
- - Upgrading wrangler 4.112.0 -> 4.123.0 - no change.
- - - Empty requirements.txt, and requirements.txt containing just "workers" - no change (this was before I found the pywrangler/pyproject.toml flow; same error persists with that flow too).
Account is on the Workers Free plan, if relevant.

### Please provide any relevant error logs

X [ERROR] A request to the Cloudflare API (/accounts/.../workers/scripts/myapp/versions) failed.

  Uncaught Error: PythonError: Traceback (most recent call last):
    File "/lib/python313.zip/_pyodide/_base.py", line 523, in eval_code
      .run(globals, locals)
       ~~~^^^^^^^^^^^^^^^^^
    File "/lib/python313.zip/_pyodide/_base.py", line 357, in run
      coroutine = eval(self.code, globals, locals)
    File "<exec>", line 9, in <module>
    File "/lib/python313.zip/_pyodide/_base.py", line 666, in pyimport_impl
      res = __import__(stem, fromlist=fromlist)
    File "/session/metadata/entry.py", line 10, in <module>
      from workers import Response, WorkerEntrypoint
    File "/lib/python3.13/site-packages/workers/__init__.py", line 3, in <module>
      raise err
  ModuleNotFoundError: No module named 'workers'
  You need to update to workers-py >= 1.90 or to pass disable_python_external_sdk

    at null.<anonymous> (pyodideRuntime-internal:emscriptenSetup:19919:14) in new_error
    at [object Object] in $wrap_exception
    at [object Object] in $pythonexc2js
    at null.<anonymous> (pyodideRuntime-internal:emscriptenSetup:22411:37) in callPyObjectKwargs
    at null.<anonymous> (pyodideRuntime-internal:emscriptenSetup:23309:20) in callKwargs
    at null.<anonymous> (pyodideRuntime-internal:emscriptenSetup:24469:87) in runPython
    at null.<anonymous> (pyodide:python-entrypoint-helper:73:17) in handleSrcImport
    at null.<anonymous> (pyodide:python-entrypoint-helper:190:17)
   [code: 10021]

mittalpk:
Fixes cloudflare/workers-py#212.

## What's wrong

`_install_requirements_to_vendor()` guessed the pyodide venv's site-packages location from `os.name` (`Lib/site-packages` on Windows, `lib/pythonX.Y/site-packages` elsewhere). That venv always targets a foreign interpreter (`cpython-*-emscripten-wasm32-musl`, see `get_uv_pyodide_interp_name()`), never the host's own — unlike `venv_workers_path` (a real native venv, which does correctly follow the host OS's convention) — so there's no single OS-conditional formula that's safe to assume holds for it.

When the guess was wrong, `uv pip install` still installed the real packages correctly (it doesn't use this guessed path at all), but they were never copied into `python_modules` — so the deployed Worker's `python_modules` bundle ended up missing the real `workers` package. At runtime, `workerd`'s Python-Workers helper (`src/pyodide/python-entrypoint-helper.ts` in `cloudflare/workerd`) deliberately injects an always-raising stub `workers` module as a safety net for exactly this situation, which is what surfaced as the reported `ModuleNotFoundError: No module named 'workers'` at Cloudflare's deploy-time validation step — not a bug in `workerd` or in `workers-runtime-sdk` itself.

Traced this the whole way from the original issue on `workers-sdk`: the traceback's `pyodideRuntime-internal`/`pyodide:python-entrypoint-helper` frames pointed at `workerd`'s own runtime code (public source), which pointed back at this repo's vendoring step as the thing responsible for actually getting the real SDK bundled.

## Fix

Adds `_find_pyodide_site_packages()`, which discovers the real site-packages directory by globbing for it (taking the shallowest match) instead of constructing a guessed path. This works regardless of what layout convention `uv` actually uses for this cross-target venv on any given host OS or `uv` version, so it doesn't just trade one unverified guess for another.

## Testing

- Built and ran the fixed CLI against a scaffold matching the reporter's exact setup (`workers-runtime-sdk` as an editable dependency, `wrangler.jsonc` with `python_workers`, an `entry.py` importing `Response`/`WorkerEntrypoint`). Confirmed the real `workers` package now correctly lands in `python_modules` after `pywrangler sync`.
- Added `test_find_pyodide_site_packages_discovers_nested_layout` (proves discovery works even at a deliberately non-standard nesting depth, not just one that happens to match a real convention) and `test_find_pyodide_site_packages_raises_when_missing` (a clear error instead of a `shutil.copytree` crash when the venv has no site-packages at all). Confirmed both fail with `AttributeError`/uncovered-path against the unpatched code.
- Full suite: 56 passed (54 existing + 2 new), 0 regressions.
- `ruff format`, `ruff check`, and `mypy` (via `uvx pre-commit run`, matching the repo's pinned tool versions) all clean.

### DIFF
--- a/packages/cli/src/pywrangler/sync.py
+++ b/packages/cli/src/pywrangler/sync.py
@@ -146,6 +146,26 @@ def create_pyodide_venv() -> None:
     run_command(["uv", "venv", str(pyodide_venv_path), "--python", interp_name])
 
 
+def _find_pyodide_site_packages() -> Path:
+    """Locate the pyodide venv's site-packages directory.
+
+    Discovered directly rather than assumed from an os.name check, since this
+    venv always targets a foreign (non-host) interpreter --
+    cpython-*-emscripten-wasm32-musl -- and there's no single formula that's
+    safe to assume holds for its internal layout on every host OS.
+    """
+    pyodide_venv_path = get_pyodide_venv_path()
+    matches = sorted(
+        pyodide_venv_path.glob("**/site-packages"), key=lambda p: len(p.parts)
+    )
+    if not matches:
+        raise RuntimeError(
+            f"Could not find a site-packages directory under {pyodide_venv_path}. "
+            "Try deleting .venv-workers and running `pywrangler sync` again."
+        )
+    return matches[0]
+
+
 def _install_requirements_to_vendor(
     plan: InstallPlan, allow_build: bool = False
 ) -> str | None:
@@ -177,11 +197,20 @@ def _install_requirements_to_vendor(
 
     # Clear pyodide venv site-packages so stale packages from previous syncs
     # don't carry over into python_modules.
-    pyv = get_python_version()
-    site_packages_path = (
-        f"lib/python{pyv}/site-packages" if os.name != "nt" else "Lib/site-packages"
-    )
-    pyodide_site_packages = get_pyodide_venv_path() / site_packages_path
+    #
+    # This venv always targets cpython-*-emscripten-wasm32-musl (see
+    # get_uv_pyodide_interp_name()), never the host's own interpreter -- unlike
+    # venv_workers_path (create_workers_venv()), which is a real native venv
+    # and does follow the host OS's own site-packages convention. Whether this
+    # *foreign*-target venv's internal layout also follows the host OS's
+    # convention, or the target's, isn't something to assume either way; it's
+    # located directly instead of guessed from os.name, which pointed at the
+    # wrong directory for at least one platform in practice (reported: real
+    # packages installed correctly by `uv`, but never copied into
+    # python_modules, so the deployed Worker fell back to workerd's
+    # always-raising stub "workers" module and failed at deploy time with
+    # ModuleNotFoundError).
+    pyodide_site_packages = _find_pyodide_site_packages()
     if pyodide_site_packages.is_dir():
         shutil.rmtree(pyodide_site_packages)
         pyodide_site_packages.mkdir()
--- a/packages/cli/tests/test_cli.py
+++ b/packages/cli/tests/test_cli.py
@@ -309,6 +309,39 @@ def test_sync_removes_stale_packages(test_dir):
     )
 
 
+def test_find_pyodide_site_packages_discovers_nested_layout(test_dir):
+    """`_find_pyodide_site_packages` must locate site-packages by discovery,
+    not by constructing a path from a guessed OS-conditional convention.
+
+    The pyodide venv always targets a foreign (non-host) interpreter, so its
+    on-disk layout isn't guaranteed to match either the POSIX or the Windows
+    native-venv convention on every host -- this was previously assumed via
+    an `os.name` check, which pointed at a nonexistent directory on at least
+    one real platform (freshly-installed packages were silently never copied
+    into python_modules, producing the reported deploy-time
+    ModuleNotFoundError). Using an unconventional depth here (rather than
+    exactly mirroring either real convention) proves the discovery is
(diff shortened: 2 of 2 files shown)


mittalpk:

ryanking13:
I don't think this is a right fix. The problem is that uv is not calculating the site-packages directory correctly in windows.

mittalpk:
You were right, and I found the concrete evidence — thank you for pushing on this.

Tested directly on Windows (uv 0.12.5): `uv venv --python cpython-3.13.2-emscripten-wasm32-musl` creates a venv with the Windows-native layout (`Lib\site-packages`, `Scripts\python.exe`), but `uv pip install --python <that venv>` crashes outright — it doesn't install to the wrong place, it fails trying to create a malformed path:

```
Using Python 3.13.2 environment at: /
error: Failed to install: click-8.4.2-py3-none-any.whl (click==8.4.2)
  Caused by: Failed to create directory `//lib/python3.13/site-packages\`
  Caused by: failed to create directory `//lib/python3.13/site-packages\`: The network path was not found. (os error 53)
```

Same result whether `--python` points at the venv directory or directly at `Scripts\python.exe`. So this PR's fix doesn't actually solve the reported bug — `_find_pyodide_site_packages()` can't discover packages that were never installed in the first place.

Filed this against uv directly: astral-sh/uv#21199. Given the real fix lives there, want me to close this PR, or is there appetite for a workaround here in the meantime (e.g. falling back to `--prefix` instead of `--python`, which a related closed uv issue confirms does work — just with host-OS layout, which the discovery approach in this PR would still handle correctly)? Happy to go either way.

ryanking13:
https://github.com/cloudflare/workers-py/pull/214 fixed it for now, so closing.

PR #214, opened 2026-08-18
fix(cli): fix calculating host python path on windows

ryanking13:
Partially rollback my change #178, with different fix. The previous change turned out it breaks `pywrangler` on windows, since when `--python` is passed, uv considers that they are in the linux environment and installs packages into incorrect directory.

This reverts to use `VIRTUAL_ENV` varaible again, but instead wipe out other env variables that can shadow VIRTUAL_ENV variables.

### DIFF
--- a/packages/cli/src/pywrangler/sync.py
+++ b/packages/cli/src/pywrangler/sync.py
@@ -25,6 +25,28 @@ from .utils import (
 
 logger = logging.getLogger(__name__)
 
+# Env vars that can cause uv to target the wrong Python environment.
+_UV_CONFLICTING_ENV_VARS = (
+    "VIRTUAL_ENV",
+    "CONDA_PREFIX",
+    "UV_PYTHON",
+    "UV_PROJECT_ENVIRONMENT",
+    "PYTHONHOME",
+)
+
+
+def _env_with_venv(venv_path: Path) -> dict[str, str]:
+    """Build a subprocess env that points VIRTUAL_ENV at *venv_path*.
+
+    Strips env vars that could redirect uv to a different environment,
+    then sets VIRTUAL_ENV explicitly.  This is preferred over ``--python``
+    for pyodide venvs because ``--python <dir>`` on Windows can cause uv
+    to install packages outside the venv's site-packages.
+    """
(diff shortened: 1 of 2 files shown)


packages/cli/src/pywrangler/sync.py
```diff
@@ -134,10 +156,32 @@ def create_workers_venv() -> None:
...
+            else:
+                logger.debug(
+                    f"Pyodide virtual environment at {pyodide_venv_path} already exists."
+                )
+                return
+        else:
+            logger.debug(
+                f"Pyodide virtual environment at {pyodide_venv_path} already exists."
+            )
+            return
```
dom96: This logic is now complex enough to live in its own function which returns `true` when the pyodide venv already exists. Doing this will also reduce duplication of the logger.debug here.

This also needs a dedicated test.

packages/cli/src/pywrangler/sync.py
```diff
@@ -25,6 +25,28 @@
...
+
+def _env_with_venv(venv_path: Path) -> dict[str, str]:
+    """Build a subprocess env that points VIRTUAL_ENV at *venv_path*.
+
+    Strips env vars that could redirect uv to a different environment,
+    then sets VIRTUAL_ENV explicitly.  This is preferred over ``--python``
+    for pyodide venvs because ``--python <dir>`` on Windows can cause uv
+    to install packages outside the venv's site-packages.
+    """
+    env = {k: v for k, v in os.environ.items() if k not in _UV_CONFLICTING_ENV_VARS}
```
dom96, line 47: Instead of excluding certain env vars, should we instead have an explicit include list? Should be more reliable. I can't think of many env vars that we would explicitly need to inherit either.

dom96 (approved):
Some comments with requested changes but accepting anyway to unblock.

packages/cli/src/pywrangler/sync.py
ryanking13: > Should be more reliable.

I am not sure about that. We cannot assume anything about the user's shell.

packages/cli/src/pywrangler/sync.py
ryanking13: Yeah, I separated it to a function and also refactored another function to use it as well.

PR #216, opened 2026-08-18
fix: fixes Request body being released too early

dom96:
It looks like we need to use create_proxy here to ensure the body isn't freed too early.

### Test Plan

```
cd packages/runtime-sdk
uv run pytest 'tests/test_fastapi.py::TestPLATFORM_IO::test_upload_spills_binary_file_to_virtual_disk[3.13]' -v
uv run pytest 'tests/test_in_workerd.py::test_in_workerd[sdk-3.13]' -v
```

### DIFF
--- a/packages/runtime-sdk/src/workers/request.py
+++ b/packages/runtime-sdk/src/workers/request.py
@@ -9,6 +9,7 @@ from .blob import Blob
 from .formdata import FormData
 from .types import FetchKwargs
 from .utils import (
+    _get_js_body,
     _is_js_instance,
     _js_headers_to_http_message,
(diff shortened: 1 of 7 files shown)


packages/runtime-sdk/src/workers/request.py
```diff
@@ -17,6 +18,23 @@
 )
 
 
+@contextmanager
+def _get_js_body(body):
```
ryanking13: We have the same function in `response.py`. Maybe let's move it to utils.py and import from there.

ryanking13 (approved):
Thanks!

PR #217, opened 2026-08-18
fix: fixes fastapi lifespan state not persisting

dom96:
Resolves issue with lifespan events https://fastapi.tiangolo.com/advanced/events/#lifespan

### Test Plan

```
$ uv run pytest 'tests/test_fastapi.py::TestPLATFORM_LIFECYCLE::test_lifespan_state_reaches_request[3.13]' -v
```

### DIFF
--- a/packages/runtime-sdk/src/workers/asgi.py
+++ b/packages/runtime-sdk/src/workers/asgi.py
@@ -49,7 +49,7 @@ def acquire_js_buffer(pybuffer):
         buf.release()
 
 
-def request_to_scope(req, env, ws=False):
+def request_to_scope(req, env, ws=False, state=None):
     from js import URL
 
     # @app.get("/example")
@@ -74,7 +74,7 @@ def request_to_scope(req, env, ws=False):
         ty = "websocket"
     else:
         ty = "http"
-    return {
+    scope = {
         "asgi": ASGI,
         "headers": headers,
         "http_version": "1.1",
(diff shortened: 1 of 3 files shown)


packages/runtime-sdk/src/workers/asgi.py
```diff
@@ -86,6 +86,10 @@ def request_to_scope(req, env, ws=False):
         "type": ty,
         "env": env,
     }
+    if state is not None:
+        # ASGI requires a shallow copy of lifespan state for each request.
+        scope["state"] = dict(state)
```
ryanking13: This would create a deep copy, no?

packages/runtime-sdk/tests/web-frameworks-test/fastapi-tests/src/worker.py
```diff
@@ -414,6 +430,12 @@ async def v1_info():
 app.include_router(v1_router)
 
 
+@app.get("/platform/lifespan-state")
+async def platform_lifespan_state(request: Request):
+    resource = request.state.platform_resource
+    return {"available": True, "closed": resource.closed}
```
ryanking13, line 436: Maybe let's also check mutating the state and see if the mutated state persists between the requests.

ryanking13 (approved):

packages/runtime-sdk/tests/web-frameworks-test/fastapi-tests/src/worker.py
dom96: I think that will fail right now, but I've got some follow up PRs on top of this which may fix it. Will keep this in mind.

PR #218, opened 2026-08-18
fix: handle ASGI WebSocket close events and environment propagation

dom96:
This fixes two gaps in ASGI WebSocket handling:
- Application-originated websocket.close messages were ignored, so clients never received the requested close code or reason.
- WebSocket scopes always received an empty env, even when the caller supplied Worker bindings.

### Test Plan

```
$ uv run pytest 'tests/test_in_workerd.py::test_in_workerd[asgi-ws-disconnect-3.13]' -v
```

### DIFF
--- a/packages/runtime-sdk/src/workers/asgi.py
+++ b/packages/runtime-sdk/src/workers/asgi.py
@@ -311,7 +311,9 @@ async def process_request(
         app_task.destroy()
 
 
-async def process_websocket(app: Any, req: "Request | js.Request") -> js.Response:
+async def process_websocket(
+    app: Any, req: "Request | js.Request", env: Any = None
+) -> js.Response:
     from js import Response, WebSocketPair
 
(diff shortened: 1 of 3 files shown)


packages/runtime-sdk/src/workers/asgi.py
```diff
@@ -362,16 +368,23 @@ async def ws_send(got):
                     server.send(jsbytes)
             if s is not None:
                 server.send(s)
-
-        else:
-            logger.warning(" == Not implemented %s", got["type"])
+            return
+        if got["type"] == "websocket.close":
+            server.close(got.get("code", 1000), got.get("reason", ""))
+            return
```
ryanking13, line 374: This partially overlaps with https://github.com/cloudflare/workers-py/pull/166, but I think we can merge this first and #166 can be rebased to handle app task ends event only (cc: @whitphx)

ryanking13 (approved):

PR #219, opened 2026-08-19
ci: fix bonk

ryanking13:
Picked up changes from https://github.com/cloudflare/workerd/pull/7050


ryanking13:
I guess this should be merged first, then the bonk will be fixed after that.

dom96 (approved):

PR #220, opened 2026-08-19
test(runtime-sdk): test hyperdrive bindings

ryanking13:
This adds tests for hyperdrive, using `pymysql` and `pg8000` packge.

Since hyperdrive tests require local mysql and postgresql server, this test runs separately from other tests, and disabled by default in local environments. It is enabled by passing `-m hyperdrive` marker when running pytest.


packages/runtime-sdk/tests/bindings-test/wrangler.jsonc
```diff
@@ -49,6 +49,21 @@
     "analytics_engine_datasets": [
         { "binding": "ANALYTICS", "dataset": "test-dataset" }
     ],
+    // "localConnectionString" points pywrangler dev at a database on the host.
+    // The databases only exist in the Hyperdrive CI job; when they are absent the
+    // dev server still starts normally and the hyperdrive suites are skipped.
```
dom96: should they fail instead of being skipped? We don't want them running if these are absent. We could end up with a broken test without being aware about it.

dom96 (approved):
Cool test!

packages/runtime-sdk/tests/bindings-test/wrangler.jsonc
ryanking13: Ah, this comment is outdated. It will not skip but fail in CI.

PR #223, opened 2026-08-21
fix(cli): remove enable date from Python 3.14

ryanking13:
reverts #197

marking it experimental again so we can bump Pyodide version and compat flag again.


dom96 (approved):

dom96:
Do we need it to be experimental again?

ryanking13:
not really. maybe let's just remove the enable date.

PR #225, opened 2026-08-21
feat: asgi.entrypoint syntax sugar for nicer asgi apps

dom96:
This changes a fastapi hello world from:

```python
from fastapi import FastAPI

app = FastAPI()

@app.get("/")
def read_root():
    return {"Hello": "World"}

import asgi
from workers import WorkerEntrypoint

        return await asgi.fetch(app, request, self.env)
```

To:

```python
from fastapi import FastAPI

app = FastAPI()

@app.get("/")
def read_root():
    return {"Hello": "World"}

import asgi
Default = asgi.entrypoint(app)
```

Also adds a websockets fastapi test.

### DIFF
--- a/packages/runtime-sdk/src/workers/asgi.py
+++ b/packages/runtime-sdk/src/workers/asgi.py
@@ -7,7 +7,7 @@ from urllib.parse import unquote
 
 import js
 
-from workers import Context, Request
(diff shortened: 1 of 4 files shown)


hoodmane (approved):
Thanks!

PR #227, opened 2026-08-25
chore(django-cf): set DJANGO_ALLOW_ASYNC_UNSAFE to 1 to make it more clear

ryanking13:
This is NFC.

Setting `DJANGO_ALLOW_ASYNC_UNSAFE` to any value allows Django to access async-unsafe code (e.g. database access) inside the event loop. Previously, it was set to `false` which gives false impression that the value is not set, but setting any value (false, true, 1, on, etc) to this env var basically enables this config, so changing it to 1 to make it more clear.

### DIFF
--- a/packages/django-cf/django_cf/__init__.py
+++ b/packages/django-cf/django_cf/__init__.py
@@ -4,7 +4,7 @@ from workers import wsgi
 
 
 async def handle_wsgi(request, app, env=None):
-    os.environ.setdefault("DJANGO_ALLOW_ASYNC_UNSAFE", "false")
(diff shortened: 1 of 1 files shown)


dom96 (approved):
Accepting because this is a no-op.

_But_ I think we shouldn't set this implicitly for our users. Ideally we shouldn't even recommend this to them if we can help it.

hoodmane:
Yeah I agree with Dominik that it isn't great to be setting this.

ryanking13:
Yeah, as we discussed internally, we should replace this with something better, but for now, this was making me (and some AI agents) keep confusing.

PR #229, opened 2026-08-26
fix: defer ASGI lifespan shutdown until request completion

dom96:
FastAPI includes features such as background tasks, streaming responses and dependencies. These can be combined in various ways and it appears that we are not handling their shutdown correctly. 

This PR ensures that we defer ASGI lifespan shutdown until streaming responses, background tasks, and dependency cleanup is complete.

### Test Plan

```
$ uv run pytest 'tests/test_fastapi.py::TestPLATFORM_LIFECYCLE' -v
```

### DIFF
--- a/packages/runtime-sdk/src/workers/asgi.py
+++ b/packages/runtime-sdk/src/workers/asgi.py
@@ -186,7 +186,7 @@ async def process_request(  # noqa: PLR0913
     # TODO(later): remove this parameter after unvendoring Python SDK from workerd
     ctx: Context | None,
     state: dict[str, Any] | None = None,
-) -> js.Response:
+) -> tuple[js.Response, Future[None]]:
     from js import Response, TransformStream
     from pyodide.ffi import create_proxy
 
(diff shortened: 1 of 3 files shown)


packages/runtime-sdk/src/workers/asgi.py
```diff
@@ -427,11 +432,30 @@ async def fetch(
...
+        await shutdown()
         raise
-    await shutdown()
+
+    if request_task.done():
+        await shutdown()
+    else:
+        from pyodide.ffi import create_proxy  # noqa: PLC0415
+
+        from workers import wait_until  # noqa: PLC0415
```
hoodmane: As a cleanup task it would be nice to add:

```py
if IN_WORKERS:
  from pyodide.ffi import create_proxy
  from workers import wait_until
```
at top level, since we are importing these in a lot of functions.

hoodmane: Probably best left to a followup though.

packages/runtime-sdk/src/workers/asgi.py
```diff
@@ -427,11 +432,30 @@ async def fetch(
...
+        async def finalize_request():
+            try:
+                await request_task
+            finally:
+                await shutdown()
+
+        finalizer_task = run_in_background(finalize_request())
+        finalizer_proxy = create_proxy(finalizer_task)
+        finalizer_task.add_done_callback(lambda _task: finalizer_proxy.destroy())
+        wait_until(finalizer_proxy)
```
hoodmane: We fixed `wait_until` so you don't have to do this:
https://github.com/cloudflare/workerd/blob/main/src/pyodide/python-entrypoint-helper.ts?plain=1#L44-L73
```suggestion
        wait_until(run_in_background(finalize_request()))
```

hoodmane (approved):
Thanks!

packages/runtime-sdk/src/workers/asgi.py
```diff
@@ -186,7 +186,7 @@ async def process_request(  # noqa: PLR0913
     # TODO(later): remove this parameter after unvendoring Python SDK from workerd
     ctx: Context | None,
     state: dict[str, Any] | None = None,
-) -> js.Response:
+) -> tuple[js.Response, Task[None]]:
```
hoodmane: I think we could type this as a `Future`, we're not using any of the extra task methods.
```suggestion
) -> tuple[js.Response, Future[None]]:
```

PR #230, opened 2026-08-27
fix(django-cf): stop using DJANGO_ALLOW_ASYNC_UNSAFE from django-cf backends

ryanking13:
Stop using discouraged `DJANGO_ALLOW_ASYNC_UNSAFE` flag and mark django-cf backends async-safe.

Since django by default wraps all the database methods with `@async_unsafe` decorator, this includes a slightly hacky solution that unwraps decorators of those methods.

### DIFF
--- a/packages/django-cf/django_cf/__init__.py
+++ b/packages/django-cf/django_cf/__init__.py
@@ -1,11 +1,7 @@
-import os
-
 from workers import wsgi
 
 
 async def handle_wsgi(request, app, env=None):
-    os.environ.setdefault("DJANGO_ALLOW_ASYNC_UNSAFE", "1")
-
     return await wsgi.fetch(app, request, env)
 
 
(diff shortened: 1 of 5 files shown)


dom96 (commented):
Hmm, isn't this effectively the same as setting `DJANGO_ALLOW_ASYNC_UNSAFE`?

ryanking13:
> Hmm, isn't this effectively the same as setting DJANGO_ALLOW_ASYNC_UNSAFE?

Partially, but a few things are different.

1. There can be other async-unsafe places, e.g. django middlewares or even another database backend (if they are using multiple databases.... though it should be very rare). This change limits the impact only to the django-cf backends to prevent unexpected side effects.

2. Since the logic it backed in to the database backend itself, users do not need to set the env variable manually. Previously, users need to set it manually or had to inherit DjangoCF class which I don't think is very optimal.

PR #231, opened 2026-08-27
fix: support persistent lifespans in asgi

dom96:
Ensures that mutable lifespans are persisted across requests. Adds fastapi and asgi-specific tests.

### DIFF
--- a/packages/runtime-sdk/src/workers/asgi.py
+++ b/packages/runtime-sdk/src/workers/asgi.py
@@ -1,8 +1,8 @@
 import logging
-from asyncio import Event, Future, Queue, create_task, ensure_future
+from asyncio import Event, Future, Queue, create_task, ensure_future, shield
 from collections.abc import Awaitable
 from contextlib import contextmanager
-from typing import Any
+from typing import Any, ClassVar
 from urllib.parse import unquote
 
 import js
@@ -323,8 +323,11 @@ async def process_request(  # noqa: PLR0913
     return response, request_task
 
 
-async def process_websocket(
-    app: Any, req: "Request | js.Request", env: Any = None
+async def websocket(
+    app: Any,
+    req: "Request | js.Request",
+    env: Any = None,
+    state: dict[str, Any] | None = None,
 ) -> js.Response:
     from js import Response, WebSocketPair
 
@@ -400,7 +403,7 @@ async def process_websocket(
 
     task = run_in_background(
         app(
-            request_to_scope(req, env if env is not None else {}, ws=True),
+            request_to_scope(req, env if env is not None else {}, ws=True, state=state),
             ws_receive,
             ws_send,
         )
@@ -427,19 +430,30 @@ async def process_websocket(
 
 
 async def fetch(
-    app: Any, req: "Request | js.Request", env: Any, ctx: Context | None = None
+    app: Any,
+    req: "Request | js.Request",
+    env: Any,
+    ctx: Context | None = None,
+    state: dict[str, Any] | None = None,
 ) -> js.Response:
+    if (req.headers.get("upgrade") or "").lower() == "websocket":
+        return await websocket(app, req, env, state=state)
+
     logger.debug("ASGI request: %s %s", req.method, req.url)
-    shutdown, state = await start_application(app)
+    shutdown = None
+    if state is None:
+        shutdown, state = await start_application(app)
     try:
         result, request_task = await process_request(app, req, env, ctx, state=state)
     except Exception:
         logger.exception("ASGI request failed")
-        await shutdown()
+        if shutdown is not None:
+            await shutdown()
         raise
 
     if request_task.done():
-        await shutdown()
+        if shutdown is not None:
+            await shutdown()
     else:
         from workers import wait_until  # noqa: PLC0415
 
@@ -447,27 +461,39 @@ async def fetch(
             try:
                 await request_task
             finally:
-                await shutdown()
+                if shutdown is not None:
+                    await shutdown()
 
         wait_until(run_in_background(finalize_request()))
 
     return result
 
 
-async def websocket(
-    app: Any, req: "Request | js.Request", env: Any = None
-) -> js.Response:
-    return await process_websocket(app, req, env)
+class AsgiWorkerEntrypoint(WorkerEntrypoint):
+    """Worker entrypoint for an ASGI application."""
+
+    app: Any
+    # WorkerEntrypoint instances are invocation-scoped, so this must live on the
+    # concrete class to preserve lifespan state across requests in the isolate.
+    _start_future: ClassVar[Future[Any] | None] = None
+
+    async def lifespan_state(self):
+        start_future = self._start_future
+        if start_future is None:
+            start_future = create_task(start_application(self.app))
+            type(self)._start_future = start_future
+        return (await shield(start_future))[1]
+
+    async def fetch(self, request):
+        state = await self.lifespan_state()
+        return await fetch(self.app, request, self.env, state=state)
 
 
-def entrypoint(app: Any) -> type[WorkerEntrypoint]:
+def entrypoint(app_: Any) -> type[AsgiWorkerEntrypoint]:
     """Create the default Worker entrypoint for an ASGI application."""
 
-    class Default(WorkerEntrypoint):
-        async def fetch(self, request):
-            if (request.headers.get("upgrade") or "").lower() == "websocket":
-                return await websocket(app, request, self.env)
-            return await fetch(app, request, self.env, self.ctx)
+    class Default(AsgiWorkerEntrypoint):
+        app = app_
 
     return Default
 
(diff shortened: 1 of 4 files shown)


packages/runtime-sdk/src/workers/asgi.py
```diff
@@ -463,11 +480,22 @@ async def websocket(
...
+        return state
+
     class Default(WorkerEntrypoint):
         async def fetch(self, request):
+            state = await lifespan_state()
             if (request.headers.get("upgrade") or "").lower() == "websocket":
-                return await websocket(app, request, self.env)
-            return await fetch(app, request, self.env, self.ctx)
+                return await process_websocket(app, request, self.env, state=state)
+            return await _fetch_request(app, request, self.env, self.ctx, state)
```
hoodmane: We shouldn't make this use private APIs, because people can't use it if they want to define a second handler in addition to fetch. We should probably make this whole body into a public function. Seems like this feature of redirecting websockets to the websocket handler would be best handled by `fetch()` as well, so how about making the existing fetch into a private helper and putting this whole body into the public fetch?

packages/runtime-sdk/src/workers/asgi.py
```diff
@@ -463,11 +480,22 @@ async def websocket(
...
+        _, state = await lifespan
+        return state
+
     class Default(WorkerEntrypoint):
         async def fetch(self, request):
+            state = await lifespan_state()
             if (request.headers.get("upgrade") or "").lower() == "websocket":
-                return await websocket(app, request, self.env)
-            return await fetch(app, request, self.env, self.ctx)
+                return await process_websocket(app, request, self.env, state=state)
```
hoodmane: Why do we have separate `websocket` and `process_websocket` functions? I think they should be combined. They don't do different things.

packages/runtime-sdk/src/workers/asgi.py
hoodmane: I suppose what we should do is make an `AsgiWorkerEntrypoint` base class. Then people can add other handlers or add extra logic around the fetch function by subclassing `AsgiWorkerEntrypoint`.
```py
class AsgiWorkerEntrypoint(WorkerEntrypoint):
    def __init__(self, ctx: Context, env: Env):
        super().__init__(ctx, env)
        self._start_future = None

    async def lifespan_state(self):
        if self._start_future is None:
            self._start_future = create_task(start_application(self.app))
        return (await self._start_future)[1]

        state = await self.lifespan_state()
        if request.headers.get("upgrade", "").lower() == "websocket":
            return await process_websocket(self.app, request, self.env, state=state)
        return await _fetch_request(self.app, request, self.env, self.ctx, state)


def entrypoint(app_: Any) -> type[WorkerEntrypoint]:
    """Create the default Worker entrypoint for an ASGI application."""

    class Default(AsgiWorkerEntrypoint):
        app = app_

    return Default
```

packages/runtime-sdk/tests/web-frameworks-test/fastapi-tests/src/test_platform_lifecycle.py
```diff
@@ -19,6 +22,26 @@ async def test_lifespan_state_reaches_request(fastapi_app):
...
+        Request("http://testserver/platform/lifespan-mutation")
+    )
+    second = await entrypoint.fetch(
+        Request("http://testserver/platform/lifespan-mutation")
+    )
+    first_data = await read_json(first)
+    second_data = await read_json(second)
+
+    assert first_data["closed"] is False
+    assert second_data == {"requests": first_data["requests"] + 1, "closed": False}
```
hoodmane:
```suggestion
    assert first_data == {"requests": 1, "closed": False}
    assert second_data == {"requests": 2, "closed": False}
```

packages/runtime-sdk/src/workers/asgi.py
dom96: Hmm, alright. This is getting a little outside the scope of this PR. But I've done it.

I wonder if we should make more of the methods in asgi.py private.

packages/runtime-sdk/src/workers/asgi.py
dom96: To be consistent with `fetch` and `process_fetch` (i.e. to avoid the public API taking the `state` as an argument)

dom96: but I'll just combine them both

packages/runtime-sdk/src/workers/asgi.py
```diff
@@ -427,47 +430,70 @@ def _close_transport(finished):
...
 
 async def fetch(
-    app: Any, req: "Request | js.Request", env: Any, ctx: Context | None = None
+    app: Any,
+    req: "Request | js.Request",
+    env: Any,
+    ctx: Context | None = None,
+    state: dict[str, Any] | None = None,
 ) -> js.Response:
+    if (req.headers.get("upgrade") or "").lower() == "websocket":
```
hoodmane, line 439:
```suggestion
    if req.headers.get("upgrade", "").lower() == "websocket":
```

packages/runtime-sdk/src/workers/asgi.py
```diff
@@ -427,47 +430,70 @@ def _close_transport(finished):
...
-    return await process_websocket(app, req, env)
+class AsgiWorkerEntrypoint(WorkerEntrypoint):
+    """Worker entrypoint for an ASGI application."""
+
+    app: Any
+    _start_future = None
+
+    async def lifespan_state(self):
+        entrypoint_type = type(self)
+        start_future = entrypoint_type.__dict__.get("_start_future")
```
hoodmane, line 480: This seems weird to me, shouldn't it be an instance attribute not a class attribute? Also, why not just:
```suggestion
        start_future = self._start_future
```

hoodmane (approved):
Generally looks reasonable, though I'd set `_start_future` on the `AsgiWorkerEntrypoint` instance rather than the class and initialize `_start_future` to `None` in `AsgiWorkerEntrypoint.__init__()`. Could adjust it in a followup if you like though.

packages/runtime-sdk/src/workers/asgi.py
```diff
@@ -427,47 +430,70 @@ def _close_transport(finished):
...
+    state: dict[str, Any] | None = None,
 ) -> js.Response:
+    if (req.headers.get("upgrade") or "").lower() == "websocket":
+        return await websocket(app, req, env, state=state)
+
     logger.debug("ASGI request: %s %s", req.method, req.url)
-    shutdown, state = await start_application(app)
+    shutdown = None
+    if state is None:
+        shutdown, state = await start_application(app)
```
ryanking13: So we don't set shutdown when state is passed?

dom96:
So we do actually want it set on the class because the instance is re-created on every worker invocation. The lifespan is supposed to stay alive for the duration of the lifetime of the isolate. I added a comment to explain this.

hoodmane:
> the instance is re-created on every worker invocation

That is really weird, can we migrate away from that?

hoodmane:
In JavaScript is the WorkerEntrypoint instantiated on every request?

dom96:
Yes, according to https://developers.cloudflare.com/workers/runtime-apis/bindings/service-bindings/rpc/#the-workerentrypoint-class

> A new instance of the class is created every time the Worker is called.

hoodmane:
Right it also says:

> Note that even though the Worker is implemented as a class, it is still stateless — the class instance only lasts for the duration of the invocation. If you need to persist or coordinate state in Workers, you should use [Durable Objects](https://developers.cloudflare.com/durable-objects).

What is this persistent lifespan state being used for? Is it volatile cache kind of stuff? Would be interested if you could post the use case.

dom96:
As far as I understand it's for reconstructible event-loop-local resources like HTTP clients, database pools, etc. Perhaps it isn't all that useful for Workers, but existing FastAPI apps that wish to run on Workers may find it useful. It's just about keeping things that are reused by each requested handled inside a single isolate.

If you don't think it's worth the complexity that's fine too. We can close :)

packages/runtime-sdk/src/workers/asgi.py
```diff
@@ -436,47 +430,70 @@ def _close_transport(finished):
...
-    app: Any, req: "Request | js.Request", env: Any = None
-) -> js.Response:
-    return await process_websocket(app, req, env)
+class AsgiWorkerEntrypoint(WorkerEntrypoint):
+    """Worker entrypoint for an ASGI application."""
+
+    app: Any
+    # WorkerEntrypoint instances are invocation-scoped, so this must live on the
+    # concrete class to preserve lifespan state across requests in the isolate.
+    _start_future: ClassVar[Future[Any] | None] = None
```
ryanking13, line 478: I think lifespan state would be more useful for Durable Objects, which has a longer lifecycle.

Currently, `DurableObject` class do not inherit `AsgiWorkerEntrypoint` so it cannot benefit from this change, so we could possibly refactor this in the future to better handle botgh regular workers and DO

ryanking13 (approved):
Approving to unblock, but probably we need to consider refactoring asgi.py as it is getting more complicated.

PR #233, opened 2026-08-27
fix: Make Headers subclass Mapping

hoodmane:
`HTTPMessage` implements all the methods needed to be considered a `Mapping` but it does not register itself as a mapping. If we try to construct a Flask response and pass an `HTTPMessage` as a header, it fails to construct because it expects either:
1. a `Mapping` (in which case it iterates over mapping.items()), or
2. an iterable of key/value pairs

If you iterate a `Mapping` directly, you only get the keys, so it doesn't work correctly because `HTTPMessage` isn't a `Mapping`.

I'm not really sure we should be using `http.client.HTTPMessage` at all here, I'd rather we had our own `Headers` class. But given we are using it I think this is a clear improvement.

### DIFF
--- a/packages/runtime-sdk/src/workers/utils.py
+++ b/packages/runtime-sdk/src/workers/utils.py
@@ -1,4 +1,5 @@
-from collections.abc import Iterator, Sequence
+import http.client
+from collections.abc import Iterator, Mapping, Sequence
 from contextlib import contextmanager
 from typing import Any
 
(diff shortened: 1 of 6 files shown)


packages/runtime-sdk/tests/workerd-test/sdk/tests/test_sdk.py
```diff
@@ -105,6 +106,15 @@ async def handler(request):
...
 
+@pytest.mark.asyncio
+async def test_fetch_response_headers_work_with_flask_response():
+    response = await fetch(
+        "https://example.com/sub_headers", headers={"Custom-Req-Header": 123}
+    )
+    flask_response = FlaskResponse(headers=response.headers)
+    assert flask_response.headers["Custom-Header-That-Should-Passthrough"] == "true"
+
+
```
ryanking13: Can we create a separate flask test suites ([simliar to django and fastapi](https://github.com/cloudflare/workers-py/tree/main/packages/runtime-sdk/tests/web-frameworks-test)) and put it there? I don't like sdk tests rely on flask package.

dom96 (approved):
Agreed with Gyeongjae, approving to unblock

PR #234, opened 2026-08-27
feat: Make pywrangler respect uv.sources from pyproject.toml

hoodmane:
And other settings. By grabbing the requirements from uv compile and then freezing them into a list and installing like that, we are losing information from pyproject.toml. We should have uv do the install from pyproject.toml so it gets it right.

1. When installing packages for Pyodide, don't use the requirement list we collected, just directly install from pyproject.toml
2. When installing native packages, use the versions we installed for Pyodide to generate a constraints file but still use pyproject.toml to get the actual requirements.


packages/cli/tests/test_resolve_sources.py
```diff
@@ -0,0 +1,242 @@
...
+    project_dir: Path,
+    local_index: str,
+    monkeypatch: pytest.MonkeyPatch,
+    *,
+    allow_build: bool = False,
+) -> Path:
+    monkeypatch.setattr(resolve_module, "get_project_root", lambda: project_dir)
+    monkeypatch.setattr(
+        resolve_module, "get_uv_pyodide_interp_name", lambda: sys.executable
+    )
```
ryanking13, line 41: It is a bit unfortunate that we need so many mocks to make these work. Maybe we can refactor these a bit later to make it better testsable.

ryanking13 (approved):
Nice! Looks much nicer than hacking toml file manually, I guess we can close https://github.com/cloudflare/workers-py/pull/107 with this.

PR #236, opened 2026-08-28
chore: Factor out some common logic in workerd tests

hoodmane:
There's a LOT of repeated junk in the various test worker.py files, this factors out some of it into a common `TestRunnerEntrypoint`. I also tried to tidy up as I did the refactoring.


packages/runtime-sdk/tests/web-frameworks-test/fastapi-tests/pyproject.toml
```diff
@@ -7,4 +7,8 @@ dependencies = [
     "python-multipart",
     "pytest",
     "pytest-asyncio<1.2.0",
+    "testlib",
 ]
+
+[tool.uv.sources]
+testlib = { path = "../testlib/dist/testlib-0.0.0-py3-none-any.whl" }
```
ryanking13, line 14: Nice, I was also thinking about what would be the best way share codes between workers, and this is pretty neat.

packages/testlib/testlib/__init__.py
ryanking13, line 1: Could you add a README file in the `testlib` directory mentioning that this is for testing and not a real package?

ryanking13 (approved):
Nice!

PR #240, opened 2026-08-31
feat(runtime-sdk): Add `serialize` option in WSGI entrypoint

ryanking13:
As we discussed internally, this adds a `serialize` option to wsgi handler so that users can serialize each request if there applications are not async safe.

Originally, I thought we should add this to `asgiref.sync.sync_to_async`, but I realized that WSGI applications don't use `asgiref.sync.sync_to_async` at all, and it is for using synchronous ORMs inside async handlers in ASGI applications. So I ended up locking the entire request.

We can probably add the same option to asgi.py as well, but asgi.py is a bit more compliated due to lifespans, so I would like to discuss and land this to wsgi first.

### DIFF
--- a/packages/runtime-sdk/src/workers/_serialize.py
+++ b/packages/runtime-sdk/src/workers/_serialize.py
@@ -1,0 +1,61 @@
+import asyncio
+from weakref import ReferenceType, WeakKeyDictionary, ref
+
+
+# Use weakref so that event loops can be garbage collected
+# when they are no longer in use.
+_locks: WeakKeyDictionary[asyncio.AbstractEventLoop, ReferenceType[asyncio.Lock]] = (
+    WeakKeyDictionary()
+)
+
+
+def _get_lock() -> asyncio.Lock:
+    """Return the serialization lock for the currently running event loop."""
+    loop = asyncio.get_running_loop()
+    lock_ref = _locks.get(loop)
+    lock = lock_ref() if lock_ref is not None else None
+    if lock is None:
+        lock = asyncio.Lock()
+        _locks[loop] = ref(lock)
+    return lock
+
+
+class RequestLock:
+    """Manage one request's ownership of the shared serialization lock."""
+
+    def __init__(self, enabled: bool) -> None:
+        self._enabled = enabled
+        self._lock: asyncio.Lock | None = None
+        self._release_on_exit = True
+
+    async def __aenter__(self) -> "RequestLock":
+        if self._enabled:
+            self._lock = _get_lock()
+            await self._lock.acquire()
+        return self
+
+    async def __aexit__(self, _exc_type, _exc, _traceback) -> None:
+        if not self._enabled:
+            return
+        if self._release_on_exit:
+            self.release()
+
+    def defer_release(self) -> None:
+        """
+        When there are background streams that need to be processed,
+        we need to defer the release of the lock until the streams are done.
+        """
+        if not self._enabled:
+            return
+        
+        self._release_on_exit = False
+
+    def release(self) -> None:
+        """Release this request's lock once."""
+        if not self._enabled:
+            return
+
+        lock = self._lock
+        if lock is not None:
+            self._lock = None
+            lock.release()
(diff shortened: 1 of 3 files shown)


packages/runtime-sdk/src/workers/wsgi.py
```diff
@@ -367,25 +374,36 @@ async def fetch(
...
+            if body is None:
+                body = await _read_body(req)
+            response = process_request(
+                app,
+                req,
+                env,
+                body,
+                on_close=request_lock.release,
+            )
+            request_lock.defer_release()
```
hoodmane, line 395: Why do we unconditionally call `defer_release()`? I guess the idea of the context manager is to release the lock on all the unsuccessful paths, but the success path will release it via the on_close method.

packages/runtime-sdk/src/workers/_serialize.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def defer_release(self) -> None:
+        """
+        When there are background streams that need to be processed,
+        we need to defer the release of the lock until the streams are done.
+        """
+        if not self._enabled:
+            return
+
+        self._release_on_exit = False
```
hoodmane, line 50: Don't we need to set this back to `True` at some point?

hoodmane: Okay I get it, RequestLock is only used once.

packages/runtime-sdk/src/workers/_serialize.py
```diff
@@ -0,0 +1,60 @@
...
+    loop = asyncio.get_running_loop()
+    lock_ref = _locks.get(loop)
+    lock = lock_ref() if lock_ref is not None else None
+    if lock is None:
+        lock = asyncio.Lock()
+        _locks[loop] = ref(lock)
+    return lock
+
+
+class RequestLock:
```
hoodmane, line 22: Perhaps this should be called `MaybeRequestLock`? We could get rid of all the `if self._enabled:` guards by doing:
```py
    def __new__(cls, enabled: bool):
        if enabled:
            return object.__new__(cls)
        return contextlib.nullcontext()
```

packages/runtime-sdk/src/workers/wsgi.py
hoodmane: Maybe could use a comment.

packages/runtime-sdk/src/workers/wsgi.py
```diff
@@ -325,11 +327,16 @@ def start_response(status, response_headers, exc_info=None):
...
-            logger.exception("Failed to close wsgi.input")
+            _close_iterable(result)
+        finally:
+            try:
+                environ["wsgi.input"].close()
+            except Exception:  # noqa: BLE001 - best-effort cleanup
+                logger.exception("Failed to close wsgi.input")
+            finally:
+                if on_close is not None:
+                    on_close()
```
hoodmane, line 339: Slightly confusing here with the nested try/finallys. So what we do is:
1. always call `_close_iterable()`, `environ["wsgi.input"].close()`, and `on_close()` in that order independently of which ones raise.
2. Errors in `environ["wsgi.input"].close()` are always suppressed (except XCPU).
3. Errors in `_close_iterable()` and `on_close()` propagate up the step
4. If both `_close_iterable()` and `on_close()` raise, the error from `on_close()`  wins.

hoodmane: It occurs to me that one general concern is if we have:
```py
try:
    do_something()
finally:
    cleanup()
```
it's possible that `do_something()` is stopped by an exceeded cpu error, and then `cleanup()` raises a different error replacing it, which gets caught and we lose track of the fact that we were supposed to stop early. This isn't really a problem introduced by this PR of course and you'd have to replace it with something awful like:
```py
run_finally = True
try:
    do_something()
except BaseException as e:
    # prevent finally block from preempting BaseExceptions
    run_finally = isinstance(e, Exception)
    raise
finally:
    if run_finally:
        cleanup()
```

hoodmane (approved):
I think the logic all makes sense but is a bit messy and confusing. I also have some lingering doubts about whether we really need this, do we have any examples of wsgi apps that break without it?

ryanking13:
> do we have any examples of wsgi apps that break without it?

Not yet. Maybe we can wait until we found a compelling example that needs this. For now people can manually use `asyncio.Lock` around the `fetch` call.

PR #241, opened 2026-08-31
fix(runtime-sdk): fix top-level entropy call for litestar and fix entropy tests

ryanking13:
It turned out entropy tests were not properly working. Since test files were not picked up during the worker initialization, they were not included during the snapshot creation and was failing to detect entropy errors.


packages/runtime-sdk/tests/test_in_workerd.py
```diff
@@ -70,14 +70,20 @@ def test_in_workerd(  # noqa: PLR0913, PLR0917  (too-many-arguments)
...
     # is only available in newer Pyodide runtimes.
-    if test_dir.name == "wsgi" and compat_date < "2026-01-01":
+    if test_dir.name == "wsgi" and python_version == "3.12":
         pytest.skip(
             "wsgi requires pyodide.ffi.run_sync (JSPI), unavailable before 2026-01-01"
         )
 
+    if test_dir.name == "entropy-patches" and python_version == "3.14":
+        pytest.skip(
+            "TODO: enable me after https://github.com/cloudflare/workerd/pull/7200 lands in wrangler"
```
dom96: shouldn't be much longer, should we just wait until this happens?

dom96 (approved):

packages/runtime-sdk/tests/test_in_workerd.py
ryanking13: yeah, I guess it is released today. let me try the new version and remove this gate.

ryanking13: Hmm, I think https://github.com/cloudflare/workerd/pull/7200 is gone out, but we still have the same issue...

```
PythonWorkersInternalError: Error loading _http_writer.cpython-314-wasm32-emscripten.so: Expected table base 14266 but got table base 14267
    at Object.getMemoryPatched (pyodide-internal:snapshot:140:19)
```

@hoodmane any idea?

ryanking13:
Merging so we can put more changes on top of this. 3.14 is still not working and some packages are found to be broken.

PR #247, opened 2026-09-07
test(runtime-sdk): test aiomysql, psycopg and asyncpg + hyperdrive

ryanking13:
Adds a smoke test for `aiomysql` and `psycopg` which are also famous Python database drivers.

### DIFF
--- a/packages/cli/src/pywrangler/utils.py
+++ b/packages/cli/src/pywrangler/utils.py
@@ -424,7 +424,7 @@ def get_pyodide_index() -> str:
         case "3.13":
             v = "0.28.3"
         case "3.14":
-            v = "314.0.2"
+            v = "314.0.6"
     return "https://index.pyodide.org/" + v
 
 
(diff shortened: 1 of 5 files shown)


packages/runtime-sdk/tests/bindings-test/src/test_hyperdrive_postgresql.py
```diff
@@ -61,6 +66,26 @@ async def test_connect(env):
     conn.close()
 
 
+@pytest.mark.asyncio
+async def test_connect_psycopg(env):
+    import psycopg  # noqa: PLC0415
+
+    assert psycopg.pq.__impl__ == "c"
+    hd = env.HYPERDRIVE_PG
+    with psycopg.connect(
```
ryanking13, line 75: psycopg support [async queries as well](https://www.psycopg.org/psycopg3/docs/advanced/async.html), but it doesn't work in Python workers. It needs more investigation.

packages/cli/src/pywrangler/utils.py
```diff
@@ -424,7 +424,7 @@ def get_pyodide_index() -> str:
         case "3.13":
             v = "0.28.3"
         case "3.14":
-            v = "314.0.2"
+            v = "314.0.6"
```
ryanking13, line 427: #246

dom96 (approved):

ryanking13:
I also added tests for asyncpg.

PR #250, opened 2026-09-09
fix(runtime-sdk): ensure finalize_request is alive inside wait_until call

ryanking13:
While working on https://github.com/cloudflare/python-workers-examples/pull/103, I noticed that finalizer_task is gc-ed and cleaned up improperly inside wait_until.

This fixes it by wrapping it with create_proxy.

Probably the long term solution is to land #90 and make `self.ctx.waitUntil` and `workers.wait_until` work exactly the same, but #90 requires JSPI which isn't available in 0.26.0a2, so it needs  more discussion.

### DIFF
--- a/packages/runtime-sdk/src/workers/asgi.py
+++ b/packages/runtime-sdk/src/workers/asgi.py
@@ -6,8 +6,9 @@ from typing import Any
 from urllib.parse import unquote
 
 import js
+from pyodide.ffi import create_proxy
 
-from workers import Context, Request, WorkerEntrypoint
+from workers import Context, Request, WorkerEntrypoint, wait_until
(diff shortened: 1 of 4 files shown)


dom96 (approved):
AI comment might have a point, but rest looks good so approving.

hoodmane:
This is a duplicate of https://github.com/cloudflare/workers-py/pull/249. I think it should be fixed in workerd:
https://github.com/cloudflare/workerd/pull/7293

ryanking13:
> This is a duplicate of https://github.com/cloudflare/workers-py/pull/249. I think it should be fixed in workerd:
https://github.com/cloudflare/workerd/pull/7293

sounds good. Thanks for fixing that

PR #251, opened 2026-09-10
feat(cli): Add --skip-native-installation option to by pass installing packages in native venv

ryanking13:
This adds a new `--skip-native-installation` flag to `pywrangler sync`.

When this flag is enabled, pywrangler will install packages only to the Pyodide venv, not in native venv. This is useful in two scenarios.

1. When there is no pre-compiled wheel for the native environment. (e.g. [`psycopg-c`](https://pypi.org/project/psycopg-c/#files))

2. In CI, when users don't need auto completion, type hints etc.

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -28,8 +28,55 @@ Then run via `uv run pywrangler`.
 
 ```bash
 uv run pywrangler --help
+# install packages
 uv run pywrangler sync
+# run dev server
+uv run pywrangler dev
+# deploy
+uv run pywrangler deploy
 ```
+
+### Advanced Usage
+
+#### Using local packages
+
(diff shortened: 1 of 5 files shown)


dom96 (approved):
Seems fine, though I wonder if having these two separate modes is the best way to solve this. The problem to me seems that your project would only have 1 or 2 such packages, so disabling native installation for the full project would affect your other packages too. Could we get pywrangler to intelligently know that a package cannot be installed natively and so skip the installation for only that package?

ryanking13:
> Could we get pywrangler to intelligently know that a package cannot be installed natively and so skip the installation for only that package?

I think that is possible.

---

But actually, it turned out that pyscopg does not need this patch (I was building the wrong package), so let me mark this as draft until we find out the actual use case of this flag.

PR #254, opened 2026-09-10
chore: Remove django lint-ignore raise inside except with no from

hoodmane:
And clean up some debug code that I noticed from the lints.

### DIFF
--- a/packages/django-cf/django_cf/db/backends/d1/base.py
+++ b/packages/django-cf/django_cf/db/backends/d1/base.py
@@ -1,4 +1,5 @@
 from django.core.exceptions import ImproperlyConfigured
+from pyodide.ffi import run_sync
 
 from ...base_engine import (
     CFDatabaseWrapper,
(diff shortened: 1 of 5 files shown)


packages/django-cf/django_cf/storage/r2.py
```diff
@@ -6,6 +6,7 @@
 from django.core.files.storage import Storage
 from django.utils.deconstruct import deconstructible
 from js import Uint8Array
+from pyodide.ffi import run_sync
```
hoodmane, line 9: There shouldn't be any case where `from js import Uint8Array` works but `from pyodide.ffi import run_sync` does not, so the guard before was pointless.

packages/django-cf/django_cf/db/backends/d1/base.py
```diff
@@ -7,6 +7,13 @@
     replace_date_trunc_in_sql,
 )
 
+try:
+    from pyodide.ffi import run_sync
+except ImportError:
+
+    def run_sync(a):
+        raise RuntimeError("Code not running inside a worker!")
+
```
hoodmane: Probably pointless to guard it like this since the other file just assumes we're in Pyodide.

ryanking13: Yeah, I guess we can just remove this guard. There was some host-side tests that were not running in workers before, but I cleaned them up.

ryanking13 (approved):
Nice!

PR #260, opened 2026-09-15
chore: Fix testlib so that failing tests are reported as failing

hoodmane:
pytest_runtest_makereport is called with status="passed", when="teardown" when teardown succeeds even when the test failed. We were previously overwriting the fail info with the teardown.


dom96 (approved):

ryanking13:
Huh, so we were silently ignoring failing tests?

hoodmane:
Indeed.

PR #263, opened 2026-09-16
chore: If WORKERD_PATH environment variable is set, use it in tests

hoodmane:
Instead of the normal process of installing workerd from npm


ryanking13 (approved):
Sounds good to me

PR #264, opened 2026-09-16
chore: Clean up redundant pytest-asyncio patches

hoodmane:
These all went into testlib, so they can be removed from all the test suites.


ryanking13 (approved):
There seems like a test failure. Otherwise looks good

PR #266, opened 2026-09-17
fix(runtime-sdk): Make NonRetryableError raised in a workflow step non-retryable

hoodmane:
The Workflows engine decides whether to retry a failed step from the JS error's
name. A Python exception reaches it as a Pyodide PythonError whose message is the
traceback, so `workers.workflows.NonRetryableError` was silently retried like any
other error.  This adds a new workflows.js shim that fixes the translation of
NonRetryableError to JavaScript so this works


packages/runtime-sdk/src/workers/utils.py
```diff
@@ -126,6 +126,43 @@ def import_from_javascript(module_name: str) -> Any:
         raise
 
 
+# Directory, relative to the worker bundle root, that `pywrangler sync` vendors Python packages
+# into. wrangler registers `.js`/`.mjs` files found under `python_modules/workers/` as ES modules
+# (everything else in `python_modules/` is opaque data), which is what lets them be imported via
+# `import_from_javascript`.
```
ryanking13, line 132: I guess these comments are a bit redundant, but probably worth mention that this is something that wrangler (workers-sdk) handles not pywrangler (https://github.com/cloudflare/workers-sdk/pull/13311).

packages/runtime-sdk/src/workers/utils.py
```diff
@@ -126,6 +126,43 @@ def import_from_javascript(module_name: str) -> Any:
...
+_SDK_JS_MODULE_PREFIX = "python_modules/workers/"
+
+
+def import_sdk_javascript_module(name: str) -> Any:
```
ryanking13: This function is not used at this point. Maybe we can introduce it when it is actually used? Also this function will not work with 0.26.0a2, so we need to be careful when using this function globally.

packages/runtime-sdk/src/workers/utils.py
```diff
@@ -144,9 +181,26 @@ def _to_python_exception(exc: JsException) -> Exception:
...
+    # runtime into a JS error named "NonRetryableError" before it reaches the
+    # Workflows engine, which is how the engine knows not to retry the step.
+    # When the engine rethrows it back to us the name either survives, or is
+    # folded into the message as a prefix.
+    if getattr(exc, "name", None) == _NON_RETRYABLE_ERROR_NAME:
+        return NonRetryableError(message)
+    if message == _NON_RETRYABLE_ERROR_NAME:
+        return NonRetryableError()
+    if message.startswith(_NON_RETRYABLE_ERROR_NAME + ": "):
+        return NonRetryableError(message[len(_NON_RETRYABLE_ERROR_NAME) + 2 :])
```
ryanking13, line 184: (No action required for this PR)

Since this error is very workflows specific, it would be nice if we can refactor it to avoid checking these branches in non-workflows entrypoints.

packages/runtime-sdk/src/workers/workflows.js
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// Extract the message the user passed to the Python exception from the
+// traceback stored in `PythonError.message`.
+function pythonExceptionMessage(e) {
+  const lines = e.message.split("\n").filter((line) => line.trim() !== "");
+  const last = lines.at(-1) ?? "";
+  const sep = last.indexOf(": ");
+  return sep === -1 ? "" : last.slice(sep + 2);
+}
```
ryanking13, line 38: This function is a little bit cryptic. Could you please add an example of input and output in the comment?

packages/runtime-sdk/tests/test_in_workerd.py
```diff
@@ -66,14 +42,27 @@ def embed(dir: Path, root: Path, level: int = 0):
         module_path = path.absolute().relative_to(module_path_root)
         embed_path = path.absolute().relative_to(root)
         if path.suffix == ".py":
-            modules.append(
-                f'(name = "{module_path}", pythonModule = embed "{embed_path}")'
-            )
+            module_type = "pythonModule"
+        elif _is_sdk_js_module(path.relative_to(dir)):
+            module_type = "esModule"
```
ryanking13, line 71: Let's make sure to deploy the workflows worker to prod before merging this PR to see if wrangler handles this properly in real production case as well.

ryanking13 (approved):
I left some comments but it looks good overall

packages/runtime-sdk/src/workers/utils.py
hoodmane: New features don't need to support 0.26.0a2 imo. Maybe we can release workers-py 2.0 and explicitly drop support for 0.26.0a2 for everything.

packages/runtime-sdk/tests/test_in_workerd.py
hoodmane: I deployed it and it works when workers-py is installed from this branch but not when it's installed from the main branch.

PR #269, opened 2026-09-18
Fix: Allow pywrangler sync to build path-based dependencies

hoodmane:
We added --allow-build but this is broken because if version 1.0 of pkg-a is
available from pyodide-index but version 1.1 is on pypyi and 1.1 has no
emscripten wheel, with --allow-build will try and fail to build 1.1 from pypi
rather than installing 1.0 from the pyodide index.

This just takes note of which packages are local paths and allows building
exactly those.


packages/cli/src/pywrangler/cli.py
```diff
@@ -136,7 +136,7 @@ def types_command(outdir: str | None, config: str | None) -> Never:
     "--allow-build/--no-allow-build",
     default=None,
     help=(
-        "Allow building source distributions and local directory sources. "
+        "Allow building any source distribution, not just local path sources. "
```
ryanking13: `not just local path sources.` part looks unnecessary, and sound like AI.

packages/django-cf/tests/in_worker/worker/pyproject.toml
```diff
@@ -4,11 +4,18 @@ version = "0.0.0"
...
 ]
 
+# `../packages` is a symlink to the monorepo's `packages/` directory that the
+# host-side `dev_server` fixture creates next to its copy of this project, so
+# the tests exercise the working-tree checkouts rather than PyPI releases.
 [tool.uv.sources]
-testlib = { path = "../testlib/dist/testlib-0.0.0-py3-none-any.whl" }
+django-cf = { path = "../packages/django-cf" }
+testlib = { path = "../packages/testlib" }
+workers-runtime-sdk = { path = "../packages/runtime-sdk" }
```
ryanking13, line 21: Nice to see that this works now!

packages/django-cf/tests/conftest.py
```diff
@@ -80,24 +78,11 @@ def _serve(project_dir: Path, tmp_path: Path) -> Generator[DevServer]:
...
-    )
-    if sync.returncode != 0:
-        pytest.fail(
-            f"pywrangler sync failed for {project_dir.name}\n{sync.stdout}\n{sync.stderr}"
-        )
-
-    # `sync` vendors the released django-cf from PyPI; tests must exercise the
-    # working tree instead.
+    # These are deployable example apps, so their pyproject.toml depends on the
+    # released django-cf from PyPI. Tests must exercise the working tree instead.
```
ryanking13, line 85: Yeah, we should split the templates and the unittests.

packages/django-cf/tests/conftest.py
```diff
@@ -154,44 +133,26 @@ def dev_server(
 ) -> Generator[str]:
     """Serve ``tests/in_worker/worker``, once per compat config.
 
-    Unlike the app fixtures above, this one runs ``uv run --no-project`` and
-    vendors the runtime SDK and django-cf working trees by hand: the worker has
-    no Django project to build, it only needs the two libraries importable.
+    Unlike the app fixtures above, the worker's pyproject.toml depends on the
+    working-tree django-cf, runtime-sdk and testlib via ``../packages/...``
+    path sources, which the ``packages`` symlink makes resolvable, so ``sync``
+    vendors them directly.
```
ryanking13: This comment now looks unnecessary.

ryanking13 (approved):
Nice!

